PR #27, opened 2021-03-19
Tim's Project 2-Beat That

ts-oh:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

**10 hours**
**Struggled again with how to create turn based for the game play.**
**Changing mode so the game could take turns.**
Comfort Level (1-5):
**3**
Completeness Level (1-5):
**3** 
**It was a challenge to use the array function in our game.**
**Well... I feel like I'm repeating code a lot and I also feel like I'm missing something or I'm overcomplicating the process of the game**

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,85 +1,68 @@
-// first initialise first player mode
-var mode = 'dice roll Mode';
+// initialise first mode asking input
+var mode = 'need input';
 
-// tracking of players
-var player = 1;
-
-// player 1 array for dice roll storage
-var rS1 = [];
-var rS2 = [];
-
-// player 2 array for dice roll storage
-var rS3 = [];
-var rS4 = [];
+// two dice array for ramdom dice roll
+var diceOne = [];
+var diceTwo = [];
 
 // player 1 and 2 number
-var p1num = '';
-var p2num = '';
+var player1Dice = [];
+var player2Dice = [];
 
 // random dice generator for two dice
-var generateDice1 = function () {
-  var diceOne = Math.floor(Math.random() * 6) + 1;
-  return diceOne;
+var generateDiceNum = function () {
+  return Math.ceil(Math.random() * 6);
 };
 
-var generateDice2 = function () {
-  var diceTwo = Math.floor(Math.random() * 6) + 1;
-  return diceTwo;
-};
+// function to concat first dice number
+function concatDiceNum1() {
+  return (diceOne[diceOne.length - 1]).toString() + (diceTwo[diceTwo.length - 1]).toString();
+}
+// function to concat second dice number
+function concatDiceNum2() {
+  return (diceTwo[diceTwo.length - 1]).toString() + (diceOne[diceOne.length - 1]).toString();
+}
 
 var main = function (input) {
-  // push random dice rolls to each array
-
-  // dice roll mode 1
+  // if input is empty, goto need input mode
   if (input == '') {
-    rS1.push(generateDice1());
-    rS2.push(generateDice2());
-    mode = 'player 1';
-    return 'Player 1' + '<br><br>' + 'Dice 1: You rolled ' + rS1[rS1.length - 1] + ' 🎲' + '<br><br>' + 'Dice 2: You rolled ' + rS2[rS2.length - 1] + ' 🎲' + '<br><br>' + 'Type either "dice1" or "dice2" to pick your first numeral of the combined number';
+    mode = 'need input';
+    return 'To play the game: input "p1" for player 1 to roll or "p2" for player 2 to roll' + '<br>' + 'you can change to player 2 after player 1 picks the dice number.';
   }
-
-  // user picks the dice
+  // if input is "p1" goto player1 mode
+  if (input == 'p1') {
+    mode = 'player1';
+    diceOne.push(generateDiceNum());
+    diceTwo.push(generateDiceNum());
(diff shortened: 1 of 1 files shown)


script.js
```diff
@@ -1,4 +1,68 @@
+// initialise first mode asking input
```
kai-rocket, line 1: From the first second- great code organisation! Clear sections with comments above each section. Very easy to read. Good job.

script.js
```diff
@@ -1,4 +1,68 @@
...
+var mode = 'need input';
+
+// two dice array for ramdom dice roll
+var diceOne = [];
```
kai-rocket, line 5: I'm confused. Why do each of these dice need to be stored in an array? Doesn't each dice just have 1 value?

script.js
```diff
@@ -1,4 +1,68 @@
...
+};
+
+// function to concat first dice number
+function concatDiceNum1() {
+  return (diceOne[diceOne.length - 1]).toString() + (diceTwo[diceTwo.length - 1]).toString();
+}
+// function to concat second dice number
+function concatDiceNum2() {
+  return (diceTwo[diceTwo.length - 1]).toString() + (diceOne[diceOne.length - 1]).toString();
+}
```
kai-rocket, line 24: Can these functions be combined into 1? E.g. `concatDiceNums(firstDiceNum)`, where `firstDiceNum` specifies whether the 1st or 2nd dice should go first?

script.js
```diff
@@ -1,4 +1,68 @@
...
-  return myOutputValue;
+  // if input is empty, goto need input mode
+  if (input == '') {
+    mode = 'need input';
+    return 'To play the game: input "p1" for player 1 to roll or "p2" for player 2 to roll' + '<br>' + 'you can change to player 2 after player 1 picks the dice number.';
+  }
+  // if input is "p1" goto player1 mode
+  if (input == 'p1') {
+    mode = 'player1';
+    diceOne.push(generateDiceNum());
```
kai-rocket, line 35: Clear logic, but why do `diceOne` and `diceTwo` need to be arrays? Would it be simpler to use var names `dice1` and `dice2`?

script.js
```diff
@@ -1,4 +1,68 @@
...
+    diceTwo.push(generateDiceNum());
+    return 'Player 1' + '<br><br>' + 'Dice 1: You rolled ' + diceOne[diceOne.length - 1] + ' 🎲' + '<br><br>' + 'Dice 2: You rolled ' + diceTwo[diceTwo.length - 1] + ' 🎲' + '<br><br>' + 'Type either "dice1" or "dice2" to pick your first numeral of the combined number';
```
kai-rocket, line 37: Is there any way we can abstract this message generation into a helper function so that we don't have to keep repeating this string in the `main` function?

script.js
```diff
@@ -1,4 +1,68 @@
...
+    return 'Your number is ' + concatDiceNum1() + '<br>' + "Input 'beat-that' to compare your number with the other player!";
+  }
+  // if user input "dice2" goto dice2 concat mode
+  if (input == 'dice2') {
+    mode = 'dice2 concat mode';
+    player2Dice.push(concatDiceNum2());
+    return 'Your number is ' + concatDiceNum2() + '<br>' + "Input 'beat-that' to compare your number with the other player!";
+  }
+  // if input is "beat-that" goto dice comparison mode
+  if (input == 'beat-that') {
```
kai-rocket, line 59: Why does the game need to wait for user input in order to proceed to the next step? I see you've been following this pattern for the past few exercises, but it's not necessary and we've discussed it in class. For future exercises, can we have the game progress modes without user input if it can?

script.js
```diff
@@ -1,4 +1,68 @@
...
+    mode = 'comparison mode';
```
kai-rocket, line 60: What's the point of setting `mode = 'comparison mode'` if there's no logic in our game that depends on the mode being "comparison mode"?

script.js
```diff
@@ -1,4 +1,68 @@
+// initialise first mode asking input
+var mode = 'need input';
+
+// two dice array for ramdom dice roll
+var diceOne = [];
+var diceTwo = [];
+
+// player 1 and 2 number
+var player1Dice = [];
+var player2Dice = [];
```
kai-rocket, line 10: 1. These var names are misleading. The names and data structures imply that they will store a list of individual dice rolls, but in fact they store a single value: the concatenated dice roll numbers. 
2. Why do we use arrays here?
3. Can we name the variables something more precise, e.g. `p1FinalNumber` or `p1CombinedNumber`?

script.js
```diff
@@ -1,4 +1,68 @@
...
+  // if user input "dice2" goto dice2 concat mode
+  if (input == 'dice2') {
+    mode = 'dice2 concat mode';
+    player2Dice.push(concatDiceNum2());
+    return 'Your number is ' + concatDiceNum2() + '<br>' + "Input 'beat-that' to compare your number with the other player!";
+  }
+  // if input is "beat-that" goto dice comparison mode
+  if (input == 'beat-that') {
+    mode = 'comparison mode';
+    if (player1Dice.slice(-1) > player2Dice.slice(-1)) {
```
kai-rocket, line 61: We do not need `p1Dice` and `p2Dice` vars to be arrays. Please add comments whenever we do relatively complex logic like `slice` to explain what and why we are doing.

kai-rocket (commented):
Good effort- keep it up! Please be clear about why we use specific data types. E.g. you didn't need to use arrays in several instances. Please also re-use code logic whenever possible, e.g. with multiple repetitive concat functions, and the generation of the user output message.

Logic needs work- please break out of the mental model that we need user input in order to change modes, and be clear why we need to use modes. e.g. changing a "mode", essentially just a global var, when there's nowhere in our code that checks for that "mode", doesn't do anything.

Please message in channel if there's anything about the feedback that's unclear! Keep it up!

PR #29, opened 2021-03-20
submitting beat-that basics assignment

ishaan-malhi:
Please fill out the survey before submitting the pull request. Thanks!

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

How many hours did you spend on this assignment? ~4hrs

Please fill in one error and/or error message you received while working on this assignment.
NaN <-- I was getting this a lot because I forgot to enter my parameters into the getResultsMessage function.

What part of the assignment did you spend the most time on?
Structuring the problem with two players, multiple game-modes and what to assign as global variable vs variable nested within a function.

Comfort Level (1-5): 2 (found it initially more challenging than previous assignment and in-class exercises.

Completeness Level (1-5): 2 (could only get basic done)

What did you think of this deliverable?
I enjoyed it. The biggest hurdle was understanding there were different game modes which wasn't immediately obvious. Once I got over that, it became a lot simpler.

Is there anything in this code that you feel pleased about? 
I think my code structure is improving, as well as my understanding of how to nest functions within functions for clean, efficient and easy-to-follow code.

PR #32, opened 2021-03-20
Beat That Game project

shayaank243:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

6-7
the formatting as well as the 
the formatting as well as trying to get the conditions to run 
Comfort Level (1-5):
2
Completeness Level (1-5):
4
it was challenging
my code is finished but not perfect

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,112 @@
-var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
+// both players names are blank
+
+var main = function (input){
+
+  var player1sRolls = [];
+  mode = 'game mode locked in';
+  // i want to express the numbers rolled into an array
+  var configurationFrDiceNumber1 = function(){
+    var displayFirstRoll = yourRandomRoll1();
+    var display2ndRoll = yourRandomRoll2();
+    console.log('rolled both dice time to facilitate the switched arrangement')
+    mode = 'dice rolled for player 1';
+    player1sRolls.push(displayFirstRoll);
+    player1sRolls.push(display2ndRoll);
+    // push both numbers into the array
+    
+  
+    
+    var changePositionForP1 = function () {
+      mode = 'change configuration of both numbers for player 1';
+      player1sRolls.pop(displayFirstRoll);
+      player1sRolls.push(displayFirstRoll);
+      console.log('arrangement changed')
+      // remove the first number, then add it back to the end of the array
+    }
+  
+
+    var bothCases = function(){
+        if (mode = 'dice rolled for player 1' && input == 'yes') {
+        mode == 'change configuration of both numbers for player 1';
+        var callChangePositioningFunc = changePosition1();
+            return 'your rolls are ' + player1sRolls;
+        } else if (input == 'no') {
+            mode == 'want to keep these numbers the same for player 1';
+            player1sRolls = [displayFirstRoll, display2ndRoll];
+            return 'your rolls are ' + player1sRolls;
+          }
+        }
+      }
+      
+
+      // repeat lines 38-69 again for player 2
+      var configurationFrDiceNumber2 = function () {
+        player2sRolls = [];
+        var displayFirstRoll = opponentRoll1();
+        var display2ndRoll = opponentRoll2();
+        // call both functions
+        player2sRolls.push(computerRandomGenNumber1);
+        player2sRolls.push(computerRandomGenNumber2);
+        // push both numbers into the array
+
+        mode = 'change configuration of both numbers for player 2';
+        // text for when the mode is about to switch
+
+        player2sRolls.pop(computerRandomGenNumber1);
+        player2sRolls.push(computerRandomGenNumber1);
+        // remove the first number, then add it back to the end of the array
+        mode = 'dice rolled for player 2';
+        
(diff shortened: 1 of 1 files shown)


shayaank243:
revamped the program to be a more structured version, project took me 10-12 hours including class time, code is officially finished and works
comfort level 2
completeness level 4
finding a way to switch the dice configuration was the hardest thing
DOUBT:
1. How can I keep a constant, never-changing variable, because i had to keep switching the variable type(player1Changed and player1KeptDaSame) for the sake of easily writing it in my if statements for when the user says yes and no.
2. I tried if, else and else if for lines 88 and 95 and it does not work, so I used 3 if functions, how would you get if, else if and else to work from lines 88 to 95

script.js
```diff
@@ -1,4 +1,97 @@
+// intialise mode in the begining
+var mode = 'player 1 mode';
+
+// create two arrays for each dice
+var player1sChanged = [];
+var player2sChanged = [];
```
kai-rocket, line 6: Var names unclear- what is "player 1s changed" supposed to mean?

script.js
```diff
@@ -1,4 +1,97 @@
...
+// this 'changed' name for this variable just denotes that i am gonna change this variable
+// create an array for the dice concat numbers
+var player1KeptDaSame = [];
+var player2KeptDaSame = [];
```
kai-rocket, line 10: Same- var names and usage unclear, even with comment above. Please communicate precisely.

script.js
```diff
@@ -1,4 +1,97 @@
...
+  }
+
+  var player1KeptDaSame = player1sChanged.concat(player1KeptDaSame);
+
+  if (input == 'no' && mode == 'dice concat mode for player 1') {
+    return 'your dice arrangement is the same, at' + player1KeptDaSame;
+  }
+  if (input == 'yes' && mode == 'dice concat mode for player 1') {
+    var temp = this.player1sChanged[0];
+    this.player1sChanged[0] = this.player1sChanged[1];
```
kai-rocket, line 41: Does this code work? What is `this.player1sChanged` and how is it different from `player1sChanged`?

script.js
```diff
@@ -1,4 +1,97 @@
...
+    mode = 'rolling dice mode for player 2';
+    console.log(mode);
+    // if in player 1 mode output randomly generated dice roll
+    player2sChanged.push(randomDiceRoll(), randomDiceRoll());
+
+    mode = 'dice concat mode for player 2';
+    console.log(mode);
+    console.log('type next after you say yes or no');
+    return 'player 2 your rolls are ' + player2sChanged + ' do you want to change the configuration of your numbers?';
+  }
```
kai-rocket, line 62: Avoid repeating logic- try to encapsulate repeated code in helper functions. See Ishaan's code for an example of this. https://github.com/rocketacademy/basics-beat-that/pull/29/files

script.js
```diff
@@ -1,4 +1,97 @@
...
+  if (input == 'yes' && mode == 'dice concat mode for player 2') {
+    var temp = this.player2sChanged[0];
+    this.player2sChanged[0] = this.player2sChanged[1];
+    this.player2sChanged[1] = temp;
+
+    mode = 'player 2s coniguration has been changed.';
+    console.log('PLAYER 2S TURN IS OVER');
+    return 'player 2,your configuration has been changed to ' + player2sChanged;
+  }
```
kai-rocket, line 77: This logic is the same as the order-choosing logic for player 1. Please avoid repeating logic and use a helper function instead with a parameter that determines if the current player is player 1 or 2

script.js
```diff
@@ -1,4 +1,97 @@
...
+    return 'congrats player 1! you win!';
+  }
+  if (player1sChanged == player2sChanged || player1sChanged == player2KeptDaSame || player2sChanged == player1KeptDaSame || player2KeptDaSame == player2sChanged) {
+    mode = 'tie';
+    return 'its a tie!!';
+  }
+  if (player2sChanged > player1sChanged || player2KeptDaSame > player1KeptDaSame || player2KeptDaSame > player1sChanged || player2sChanged > player1KeptDaSame) {
+    mode = 'player 2 wins';
+    return 'congrats player 2! you win!';
+  }
```
kai-rocket, line 95: 1. No need to store "Changed" and "KeptDaSame" variable names for each player. Can just store a "FinalNumber" var for each player, regardless of whether the dice order was changed or kept the same.
2. This comparison logic can be performed in a helper function to de-clutter `main`.

kai-rocket (commented):
Does this work? Great work getting all the logical components in there!

kai-rocket:
> How can I keep a constant, never-changing variable, because i had to keep switching the variable type(player1Changed and player1KeptDaSame) for the sake of easily writing it in my if statements for when the user says yes and no.
Keep a global called `player1FinalNum` and assign the final number to that variable after the user chooses whether they want the 1st or 2nd dice first.

> I tried if, else and else if for lines 88 and 95 and it does not work, so I used 3 if functions, how would you get if, else if and else to work from lines 88 to 95
If you return from inside an if statement, eslint will prevent you from using `else` after that if statement because the else will be implicit. What you did is ok.

PR #40, opened 2021-06-08
Project2: Beat That [ck - Section 1]

makck:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

1hr 15mins

Undefined variables.

Coming up with the main logic of the game.

Comfort Level (1-5):
3

Completeness Level (1-5):
3

Nice way of introducing arrays! :)

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,70 @@
+// Global variable to track the game mode, ie. if it's player 1's or player 2's turn.
+var gameMode = "1";
+
+// Global variable to track the player's number selections.
+var player1Final = "";
+var player2Final = "";
+var player1Temp = [];
+var player2Temp = [];
+
+// Random number generator between 1 to 6
+var generateDiceNumber = function () {
+  return Math.floor(Math.random() * 6) + 1;
+};
+
+// Function to roll 2 dices and assign them to an array.
+var diceRoll = function () {
+  var dice1 = generateDiceNumber();
+  var dice2 = generateDiceNumber();
+  var diceCombi = [dice1, dice2];
+  return diceCombi;
+};
+
+// Function to display message for each roll from player 1 and player 2. Function takes an array as a parameter.
+var displayRoll = function (playerRoll) {
+  var playerNum = `Welcome Player 1.`;
+  if (gameMode == 2) {
+    playerNum = `Welcome Player 2.`;
+  }
+  return `${playerNum} <br> You rolled ${playerRoll[0]} for Dice 1 and ${playerRoll[1]} for Dice 2. <br> Input the number of the dice to determine the first numeral.`;
+};
+
+// Function takes in player seleted dice number, and sets it as the first numeral. Reads 2 parameters, user input and array with dice numbers.
+var sequenceNumber = function (userNum, diceNum) {
+  return "" + diceNum[userNum - 1] + diceNum[diceNum.length - userNum];
+};
+
+// Game logic to check for the larger number between player 1 and 2's selections and print winning player message.
+var compareNumbers = function (player1, player2) {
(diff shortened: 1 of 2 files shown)


makck:
The "More Comfortable" section was done based on original question which required players to input their own sequencing using zero-indexing.

script.js
```diff
@@ -1,4 +1,204 @@
...
+  }
+  return playerDiceRoll;
+};
+
+// Get player to select the sequence of the dice. Function takes in the sequence input, and the array of dice rolls.
+var getSequence = function (input, diceArray) {
+  var combinedNum = 0;
+  var autoCombinedNum = 0;
+  var indexArray = input.split("").map(Number);
+  // This loop takes in the user input sequence and generate the combined number.
```
eddiejpot, line 61: Interesting use of the map method. Could you explain what this line of code (line60) does?

script.js
```diff
@@ -1,4 +1,204 @@
...
+  for (var i = 0; i < indexArray.length; i += 1) {
+    combinedNum = combinedNum * 10 + diceArray[indexArray[i]];
+  }
```
eddiejpot, line 64: Great work using for loops!

script.js
```diff
@@ -1,4 +1,204 @@
...
+        } else {
+          index += 1;
+        }
+      }
+      autoCombinedNum = autoCombinedNum * 10 + maxNum;
+      tempArray.splice(maxIndex, 1);
+      maxNum = 7;
+      index = 0;
+    }
+  }
```
eddiejpot, line 101: Since you've started using for loops. It might be cleaner to use them here instead of while :) 
Same goes for the rest of the while loops

script.js
```diff
@@ -1,4 +1,204 @@
...
+          } <br> You have formed the number ${
+            thisPlayerNum[0]
+          } <br> The auto generated ${gameVersion} number is ${
+            thisPlayerNum[1]
+          }.  <br><br> Click "submit" for the next player to roll the dice.`;
+          playerCount += 1;
+        }
+        return outputValue;
+      }
+    }
```
eddiejpot, line 194: Interesting use of the while loop ! Very nice. Great that it works!

script.js
```diff
@@ -1,4 +1,204 @@
...
-  return myOutputValue;
+  if (gameVersion == "") {
+    return getVersion(input);
+  } else if (numPlayers == 0) {
+    return getPlayers(input);
+  } else if (numPlayers != 0 && numDice == 0) {
+    return getNumDice(input);
+  } else {
+    while (playerCount < numPlayers) {
+      if (gameMode == 1) {
```
eddiejpot, line 158: Good use of if and else ifs  

Usually when there are a number of else ifs we'll try to simplify by using if and return . Reason being that it makes the code more readable. Something like this
```
  if (gameVersion == "") {
    return getVersion(input);
  } 
  
  if (numPlayers == 0) {
    return getPlayers(input);
  } 
  
  if (numPlayers != 0 && numDice == 0) {
    return getNumDice(input);
  } else {
    while (playerCount < numPlayers) {
      if (gameMode == 1) {  ..................
```

script.js
makck: haha I took this off the hint in the initial project question, before Kai updated it to remove the "Get user to input sequence of dice" requirement. 

As I understand though, the user input is a string, the split method will then break up each character in the string since the delimiter is "". The map method then applies a function to all elements that have been split up. In this case, it applies the Number function, which converts all elements to an integer data type in the array. 

This is necessary so we can make use of each element in IndexArray to act as an index for the diceArray array in line 63.

PR #54, opened 2021-06-15
Beat That  - Justin Wong

JustinWong98:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

1 on base.

About 5-6 hours after everything.
I had some trouble making the leaderboard work for the more comfortable section - especially when having to deal with different array sizes if the number of players changed in-between rounds.

Variable number of dice/players - encountered array related errors while trying to implement total scores. I think I also spent too much time trying to make things work using just what we have learnt in the curriculum rather than look online for different methods to solve the problem. I will be doing that for my next project

Comfort Level (1-5):
3.5
Completeness Level (1-5):
4.5

It was sufficiently challenging!

I was able to come up with some solutions to problems using arrays. I was able to combine all the modes together with more ease than I expected after suffering through some previous parts.

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,81 @@
+var mode = "player1start";
+var combinedDiceRoll1 = 0;
+var combinedDiceRoll2 = 0;
+var outcome1 = 0;
+var outcome2 = 0;
+var p1Choice = 0;
+var p2Choice = 0;
 var main = function (input) {
-  var myOutputValue = 'hello world';
+  if (mode == "player1start" || mode == "player2start") {
+    var rollDice1 = diceroll();
+    var rollDice2 = diceroll();
+    outcome1 = rollDice1 * 10 + rollDice2;
+    outcome2 = rollDice2 * 10 + rollDice1;
+    console.log(outcome1);
+    console.log(outcome2);
+    var myOutputValue =
+      `You have rolled ${rollDice1} and ${rollDice2}! Type either 1 or 2 to pick your combined number: ` +
+      "<br>" +
+      `1: ${outcome1}` +
+      "<br>" +
+      `2: ${outcome2}`;
+    if (mode == "player1start") {
+      mode = "player1pick";
+    } else {
+      mode = "player2pick";
+    }
+    return myOutputValue;
+  } else if (mode == "player1pick" || mode == "player2pick") {
+    if (input == 1) {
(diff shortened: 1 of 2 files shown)


script.js
```diff
@@ -1,4 +1,346 @@
...
+    var knockoutModeOutput = setKnockoutMode(input);
+    return knockoutModeOutput;
+  }
+  if (mode == "gameModeSelect") {
+    var gameModeOutput = setGameMode(input);
+    return gameModeOutput;
+  } else {
+    var diceRolls = multipleDiceRoll(noOfPlayers, noOfDice, input);
+    return diceRolls;
+  }
```
emily-zhangg, line 34: good job setting the logic for each game mode into individual functions. It would be great if you could place the individual helper functions outside of the main function too!

script.js
```diff
@@ -1,4 +1,346 @@
...
+  if (mode == "gameMode") {
+    return `Please submit to roll the dice.`;
+  }
+  if (mode == "selectMode") {
+    return `Player ${
+      currentPlayer + 1
+    }, please select your die in order starting from 0.`;
+  } else if (mode == "faceOff") {
+    return "The game has ended!";
```
emily-zhangg, line 344: since you have multiple cases, you could consider using a switch statement instead (https://www.w3schools.com/js/js_switch.asp). However good job using if statements to implement this logic

yiqiingg (commented):
great job getting this far in the project!! I like the use of arrays in your code and the sectioning of your game modes logic into helper functions

script.js
```diff
@@ -1,4 +1,346 @@
...
+      }
+      myOutputValue = `Player ${
+        currentPlayer + 1
+      },You have rolled ${diceRollsArray}! <br> 
+      Enter the order of the dice! The dice order starts from 0.`;
+      mode = "selectMode";
+      return myOutputValue;
+    } else if (mode == "selectMode") {
+      indexesArray = input.split("").map(Number);
+      combinedNum = 0;
```
emily-zhangg, line 161: This is the first time that you have used combinedNum, so it would be clearer if you declared it as a new variable so that it is clearer. JS will create a new variable if it has not seen the variable name before but we would want to declare all our new variable for clarity

PR #98, opened 2021-07-29
Jessica Liurensia, Basics-9-1

JessicaLiurensia:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

How many hours did you spend on this assignment? 5-6 hours

the first error is the undefined dice, after i pick the order of the dice.
the second error which i spend most of the time figuring it out, is because the game doesn't restart and it keeps capturing the same arrays [0] and [1]. because i write ${player1DiceRolls[0]}${player1DiceRolls[1]} or vice versa. so i manage to restart the game without having to click refresh.

trying to solve the problem with the code array that i write. but manage to find another solution to it. and also trying to refine the repeating codes

Comfort Level (1-5): 3

Completeness Level (1-5): 4

What did you think of this deliverable? better than the first one i make

Is there anything in this code that you feel pleased about? i manage to refine it, and reuse the same code again


script.js
```diff
@@ -1,4 +1,294 @@
...
+// restarting the game
+var restartGame = function () {
+  currentGameMode = "player 1 dice roll";
+  currentPlayer = 1;
+  player1DiceRolls = [];
+  player2DiceRolls = [];
+  // pickDiceOrder;
+  player1CombineNumber = 0;
+  player2CombineNumber = 0;
+};
```
eddiejpot, line 192: Nice use of a function to restart the game!

script.js
```diff
@@ -1,4 +1,294 @@
...
+    counter += 1;
+  }
+
+  // to find the sum of player 2 scores
+  player2TotalScores = 0;
+  var counter = 0;
+  while (counter < player2Scores.length) {
+    player2TotalScores = player2TotalScores + player2Scores[counter];
+    counter += 1;
+  }
```
eddiejpot, line 240: Nice use of while loop in this :)!

script.js
```diff
@@ -1,4 +1,294 @@
...
+    var myOutputValue = playerRollDice();
+  } //player 2 choose the order of the dice
+  else if (currentGameMode == "player 2 choose dice order") {
+    var myOutputValue = autoGenerateDice();
+  }
+
+  console.log("player 1 scores: " + player1Scores);
+  console.log("player 2 scores: " + player2Scores);
+
   return myOutputValue;
```
eddiejpot, line 293: Great use of functions to help simplify the logic in the main! Nicely done!

script.js
```diff
@@ -1,4 +1,294 @@
...
+
+// 2 mode: "normal" (highest combined number win) || "reverse" (lowest combined number win)
+
+var currentGameMode = "waiting for user to pick the mode";
+var mode;
+var currentPlayer = 1;
+var player1CombineNumber = 0;
+var player2CombineNumber = 0;
+var player1TotalScores = 0;
+var player2TotalScores = 0;
```
eddiejpot, line 14: Whenever we have more than 1 player in a game, it's usually good to store the values in a data type that can hold multiple values (Just in case there will e multiple players next time) 

So my variables might look something like this:
Can either use an array or object 
```
// Array method 
// index 0 is playerOne and index 1 is player 2 
var playersTotalScore= [10,8]

//Object method
var playersTotalScore= {
    1: 10,
    2: 8
}
```

eddiejpot (commented):
Good job! Looking forward to the final project!

PR #102, opened 2021-07-30
Josephina (Section 9-7) Beat That Project

phinafung:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

How many hours did you spend on this assignment? 10

Please fill in one error and/or error message you received while working on this assignment. function not defined

What part of the assignment did you spend the most time on? first part 

Comfort Level (1-5): 3

Completeness Level (1-5): 4

What did you think of this deliverable? cool but tiring

Is there anything in this code that you feel pleased about? -

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,340 @@
-var main = function (input) {
-  var myOutputValue = 'hello world';
+// Variables to store player 1 and player 2's scores:
+var P1TotalScore = 0;
+var P2TotalScore = 0;
+
+// To track game modes & statuses
+var MODE_CHOOSE_GAME_MODE = `WAITING_FOR_USER_TO_CHOOSE_GAME_MODE`;
+var EITHER_HIGHEST_OR_LOWEST_WINS = `EITHER HIGHEST OR LOWEST WINS`;
+var MODE_HIGHER_NUMBER = `higher number`;
+var MODE_LOWER_NUMBER = `lower number`;
+var MODE_PLAYER_1_ROLL_DICE = `WAITING_FOR_PLAYER_1_TO_ROLL_DICE`;
+var MODE_PLAYER_1_PICK_ORDER = `WAITING_FOR_PLAYER_1_TO_PICK_DICE_ORDER`;
+var MODE_PLAYER_2_ROLL_DICE = `WAITING_FOR_PLAYER_2_TO_ROLL_DICE`;
+var MODE_PLAYER_2_PICK_ORDER = `WAITING_FOR_PLAYER_2_TO_PICK_DICE_ORDER`;
+var MODE_WINNER = `SHOW_WINNER`;
+
+//Current game mode
+var mode = MODE_CHOOSE_GAME_MODE;
+var theModeUserInput = EITHER_HIGHEST_OR_LOWEST_WINS;
+
+// Function for leaderboard for Mode: Higher number wins:
+var leaderBoardHigher = function () {
+  if (Number(P1TotalScore) > Number(P2TotalScore)) {
+    myOutputValue = `Player 1<br>Player 2`;
+  } else if (Number(P2TotalScore) > Number(P1TotalScore)) {
+    myOutputValue = `Player 2<br>Player 1`;
+  } else if (Number(P1TotalScore) == Number(P2TotalScore)) {
+    myOutputValue = `Draw`;
+  }
   return myOutputValue;
 };
+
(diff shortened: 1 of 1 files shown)


script.js
```diff
@@ -1,4 +1,340 @@
...
+// To track game modes & statuses
+var MODE_CHOOSE_GAME_MODE = `WAITING_FOR_USER_TO_CHOOSE_GAME_MODE`;
+var EITHER_HIGHEST_OR_LOWEST_WINS = `EITHER HIGHEST OR LOWEST WINS`;
+var MODE_HIGHER_NUMBER = `higher number`;
+var MODE_LOWER_NUMBER = `lower number`;
+var MODE_PLAYER_1_ROLL_DICE = `WAITING_FOR_PLAYER_1_TO_ROLL_DICE`;
+var MODE_PLAYER_1_PICK_ORDER = `WAITING_FOR_PLAYER_1_TO_PICK_DICE_ORDER`;
+var MODE_PLAYER_2_ROLL_DICE = `WAITING_FOR_PLAYER_2_TO_ROLL_DICE`;
+var MODE_PLAYER_2_PICK_ORDER = `WAITING_FOR_PLAYER_2_TO_PICK_DICE_ORDER`;
+var MODE_WINNER = `SHOW_WINNER`;
```
jeremylim91, line 14: Good use of variables to store all the string connstants 💪🏻

script.js
```diff
@@ -1,4 +1,340 @@
...
+  } else if (Number(P2TotalScore) > Number(P1TotalScore)) {
+    myOutputValue = `Player 2<br>Player 1`;
+  } else if (Number(P1TotalScore) == Number(P2TotalScore)) {
+    myOutputValue = `Draw`;
+  }
+  return myOutputValue;
+};
+
+// Function for leaderboard for Mode: Lower number wins:
+var leaderBoardLower = function () {
```
jeremylim91, line 33: We could try having a more semantic name for `leaderBoardLower` .  Additionally, in general, a good practice is to use  a verb as the first word in function names. like  **calculate**TotalSum, or **get**AllDice.

script.js
```diff
@@ -1,4 +1,340 @@
...
+  var leaderBoardRanksHigher = leaderBoardHigher();
+  var leaderBoardRanksLower = leaderBoardLower();
+
+  // User to choose game mode
+  if (mode == MODE_CHOOSE_GAME_MODE) {
+    console.log(mode);
+    var myOutputValue = `Hello! Welcome to the game of BEAT THAT.<br><br>Please select your game mode preference.<br><br>Choose either "higher number" or "lower number". <br><br>Good luck!!`;
+    mode = MODE_PLAYER_1_ROLL_DICE;
+    console.log(mode);
+  } else if (input == `higher number`) {
```
jeremylim91, line 56: Good use of `else if`  to 1. assert control over code flow, and 2. highlight to readers (like myself) that there is a mutually exclusive relationship between these blocks of code (i.e. only either one of them can be true).

script.js
```diff
@@ -1,4 +1,340 @@
...
+
+  // Player 1's turn
+  else if (
+    mode == MODE_PLAYER_1_ROLL_DICE &&
+    (theModeUserInput == MODE_HIGHER_NUMBER ||
+      theModeUserInput == MODE_LOWER_NUMBER)
+  ) {
+    console.log(mode);
+    console.log(theModeUserInput);
+    P1DicePermutation1 = P1RandomDiceRoll1 * 10 + P1RandomDiceRoll2;
```
jeremylim91, line 76: Cool mathematical approach! An alternative is to treat the two numbers as a string, concaten\ate them, then cast them into a number:

```
//concat 2 strings
var concatenatedNum= "3"+ "5"

//Cast the string into a number 
var number = Number(concatenatedNum)
```
See [here](https://www.w3schools.com/jsref/jsref_number.asp) for more info on `Number()

jeremylim91 (commented):
Hey Josephina, good effort! The code continues to reflect that you are keeping up well with class material. I see that you've made effort to try an abstract logic out of `main`  by using helper functions, and that's definitely an improvement from your previous PR! Do continue to abstract more code where possible; it becomes easier as you get the hang of it.

At the same time, keep in mind that the idea is not to have a main that is totally devoid of code--the idea is to have a main that is **concise** , meaning that I can read thru `main` and very quickly get a gist of your code design, without having to grapple with implementation details. 

Good job, jia you!💪🏻

PR #120, opened 2021-07-31
v1 - Game Only

Melvin-Disastra:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

3 Hours
Unreachable code
Syntax
Comfort Level (1-5):
2
Completeness Level (1-5):
2
Hard to troubleshoot. Game rules complicated to implement in code.
Not yet. Still need to finish more comfortable sections


script.js
```diff
@@ -1,4 +1,128 @@
-var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
-};
+// Game states
+//var gameState_diceCount = "gameState_diceCount";
+var gameState_diceRoll = "gameState_diceRoll";
```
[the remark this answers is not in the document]
lightweightcoder: Please name your variables in camelCase. Reason is that underscores (aka. snake_case) is used for defining constants, which we will not cover in this course.

script.js
```diff
@@ -1,4 +1,128 @@
...
+var player2Rolls = [];
+
+// Dice simulation
+var getDiceRoll = function () {
+  return Math.ceil(Math.random() * 6);
+};
+var getDiceRoll2 = function () {
+  var twoDiceRolled = [getDiceRoll(), getDiceRoll()];
+  // Insert round results to player results tracker array
+  if (currPlayer === 1) {
```
lightweightcoder, line 26: Good usage of triple equals!

script.js
```diff
@@ -1,4 +1,128 @@
...
+    player2Num = playerNum;
+    return playerNum;
+  }
+};
+
+function winner() {
+  if (player1Num > player2Num) {
+    return (winner = "Player 1");
+  } else {
+    return (winner = "Player 2");
```
lightweightcoder, line 66: remember to consider tie condition.

script.js
```diff
@@ -1,4 +1,128 @@
...
+
+var playerNumber = function () {
+  var diceArray;
+  if (currPlayer === 1) {
+    diceArray = player1Rolls;
+  }
+  if (currPlayer === 2) {
+    diceArray = player2Rolls;
+  }
+  return diceArray;
```
lightweightcoder, line 47: Because you wrote a return statement in line 47, your code will always exit from the `playerNumber` function. This means that lines 48 to 59 will never run. Try removing the return so you will be able to run the code below.

lightweightcoder (commented):
Overall, just keep in mind that we should name our variables in camelCase and that the `return` statement will exit the function it is in. Also, you can try to `console.log` at places where you are not sure if your code will reach or when you want to know the value of a variable at that line.

PR #131, opened 2021-07-31
9-7_Jia Hui_project 2

jaihiu:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

Comfort Level (1-5):

Completeness Level (1-5):

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,23 +1,89 @@
+var PlayerOne = [];
+var playerTwo = [];
+
 // Game modes
-var gameModeDiceRoll = "game mode dice roll";
+var gameModeDiceRoll = "game mode dice roll player 1";
 var gameModeChooseDice = "game mode choose dice";
 
-var gameMode = gameModeDiceRoll;
+// Game mode 1 roll dice
+var mode = gameModeDiceRoll;
 
 // generate dice number from 1-6
 var diceRoll = function () {
   var randomDiceRoll = Math.ceil(Math.random() * 6);
   return randomDiceRoll;
 };
-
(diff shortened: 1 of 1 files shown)


script.js
```diff
@@ -1,4 +1,89 @@
+var PlayerOne = [];
+var playerTwo = [];
+
+// Game modes
+var gameModeDiceRoll = "game mode dice roll player 1";
```
jeremylim91, line 5: Good use of variable names to store the string constants

script.js
```diff
@@ -1,4 +1,89 @@
...
+        playerOne[1] +
+        playerOne[0] +
+        ". It is now Player 2's turn.";
+      mode = gameModeDiceRoll;
+      return myOutputValue;
+    }
+  }
+
+  // Player 2
+  if (mode == gameModeDiceRoll) {
```
jeremylim91, line 54: I think you've realised that the game doesn't work the way you think it would? One key reason is that the mode name for the first player is the same as the mode name for the second player. This is one area that needs to be fixed, so that we can clearly separate what values are assigned to the first player and what should be assigned to the second player.

jeremylim91 (commented):
Hey Jia Hui! Good effort on this, can telll you put thought into it. You've prolly realised that  the game doesn't flow the way you want it to. One key reason is that there should be separate modes for when player 1 is playing, and then another for when player 2 is playing. Can also consider using more `else if` statements to have greater control of the flow of the code :)

PR #166, opened 2021-09-10
Section 1 Sim Wang: Beat that v1

Simwang94:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

>5
variable not defined
Using loops to decide on the winning number
Comfort Level (1-5):
2
Completeness Level (1-5):
3
It works, but I don't have time to build the tracking of winning stats and leaderboard, as well as reverse game mode.
Loop logic to decide on winning number and if there is a draw

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,191 @@
+//Array to store each round of dice roll
+var randomRollArray = [];
+
+//Array to store player names
+var playerNameArray = [];
+
+//Array to store each player's biggest number
+var playerBiggestNumberArray = [];
+
+//Starting game mode - Enter number of players
+var gameMode = "Enter number of players";
+//gameMode 2 = 'Roll dice'
+
+var playerTurn = 0;
+
 var main = function (input) {
-  var myOutputValue = 'hello world';
+  //Storing of player names and starting game
+  if (gameMode == "Enter number of players" && input == "start") {
+    gameMode = "Roll dice";
+    return (myOutputValue =
+      "The game begins. Please follow the instructions below:<br><br>Player " +
+      playerNameArray[0] +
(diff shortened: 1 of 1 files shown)


script.js
```diff
@@ -1,4 +1,191 @@
...
+
+//Array to store player names
+var playerNameArray = [];
+
+//Array to store each player's biggest number
+var playerBiggestNumberArray = [];
+
+//Starting game mode - Enter number of players
+var gameMode = "Enter number of players";
+//gameMode 2 = 'Roll dice'
```
eddiejpot, line 12: Great work with the comments to describe the global variables. 

I would however declare all the possible gameModes here too. Reason being that it keeps the code cleaner as we have a quick overview of all the gameModes, if we were to change a gameMode, it would be easier as well. 

Something like 
```
var gameMode1 = "Enter number of players";
var gameMode2 = "Roll dice";
var currentGameMode = gameMode1;
```

script.js
```diff
@@ -1,4 +1,191 @@
...
+    playerName +
+    ": You rolled " +
+    randomRollArray[0] +
+    " and " +
+    randomRollArray[1] +
+    ".<br><br>Your biggest number is " +
+    generateBiggestNumber();
+
+  return myOutputValue;
+};
```
eddiejpot, line 143: Nice work extracting this function out from the main logic. 
The next improvement would be to ensure that the function naming reflects what it does. 
E.g. this function is called generateNumber, I would expect it to return a number, however it returns a sentence/string.  

Alright for smaller apps, but in the future when we create bigger apps we would want our functions to be named clearly :)

PR #184, opened 2021-10-14
Kenneth Ong (basics-11-8)

kennethongcs:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

20

Multiply logical errors and assignment errors due to a large amount of variables.

Determining the game logic

Comfort Level (1-5):
3
Completeness Level (1-5):
4

Great! Helped to expand on what I have learned from past lessons.

Getting the logic done. (from player 1 rolling the dice, to player 2 rolling the dice and recording both scores)


script.js
```diff
@@ -1,4 +1,133 @@
...
+const btnDice2 = document.querySelector('.select-dice-2');
+const inputEl = document.querySelector('#input-field');
+const outputEl = document.getElementById('output-div');
+const instructionsEl = document.getElementById('instructions');
+const imgDice1 = document.querySelector('.dice1');
+const imgDice2 = document.querySelector('.dice2');
+const currentScoreEl0 = document.getElementById('current-0');
+const currentScoreEl1 = document.getElementById('current-1');
+const playerScoreEl0 = document.querySelector('.score-0');
+const playerScoreEl1 = document.querySelector('.score-1');
```
leechuanxin: Firstly, good work on tampering with the HTML file and creating your own HTML / DOM elements! Pretty nice UI, might I add, such a simple change but it makes the familiar but boring starter pack more refreshing.

I noticed you mentioned `Multiple logical errors and assignment errors due to a large amount of variables` under your difficulties in your personal review. From what I can see, you do not seem to have a lot of variables created and scoped within the functions. Most of them are constant variables in the global scope, which is fine. This is given that you are manipulating a lot of DOM elements, which may need access within some functions (for changing the CSS classes), or assigning event listeners (outside of functions).

For starters, I will work with cleaning up the variables you are `querySelect`-ing. Let's start with consistency. As you may have realised in [line 7](https://github.com/rocketacademy/basics-beat-that/blob/ea92b1cd971ce3348831bb7555c925c772b0b469/script.js#L7), you can also `.querySelector` by the IDs of the DOM elements. Although it will not affect the performance of your game, I will stick to using `.querySelector` throughout the application for selecting IDs too, by changing the rest of the `.getElementById` in [lines 8 to 9](https://github.com/rocketacademy/basics-beat-that/blob/ea92b1cd971ce3348831bb7555c925c772b0b469/script.js#L8-L9), and [lines 12 to 13](https://github.com/rocketacademy/basics-beat-that/blob/ea92b1cd971ce3348831bb7555c925c772b0b469/script.js#L12-L13).

There may also be elements `querySelect`-ed that are unused right now. [inputEl](https://github.com/rocketacademy/basics-beat-that/blob/ea92b1cd971ce3348831bb7555c925c772b0b469/script.js#L7) is one example, I don't see it being used anywhere else besides [a line that has since been commented out](https://github.com/rocketacademy/basics-beat-that/blob/ea92b1cd971ce3348831bb7555c925c772b0b469/script.js#L106). I understand that you may have created it in lieu of tackling the other More Comfortable requirements, and you can ignore this comment if that's the case.

For certain variables, it might be better to have them scoped within the functions, if they are only intended to be used in certain functions. `outputEl`, `currentScoreEl0`, and `currentScoreEl1` are only used within the `rollDice` function, so they can be scoped within the function like so:

```
const rollDice = function () {
  const outputEl = document.querySelector("#output-div");
  const currentScoreEl0 = document.querySelector("#current-0");
  const currentScoreEl1 = document.querySelector("#current-1");
  ...
}
```

Similarly, `playerScoreEl0` and `playerScoreEl1` can be scoped within the `determineWinner` function, since they are only used there.

script.js
```diff
@@ -1,4 +1,133 @@
...
+let currentRoll = [];
+let nameOfPlayers = ['Player 1', 'Player 2'];
+
+// Main instructions
+instructionsEl.textContent = `${nameOfPlayers[activePlayer]}, please roll dice to start the game! 🎲`;
+
+// determines winner based on the current scores
+const determineWinner = function () {
+  if (currentRoll[0] === currentRoll[1]) {
+  } else if (currentRoll[0] > currentRoll[1]) {
```
leechuanxin: I probably wouldn't use empty `if` blocks. Sure it works, but it's not considered good practice. This is because it may cause confusion when code reading (for other developers), and most developers will assume it implies incomplete refactoring.

You can do 
```
if (currentRoll[0] < currentRoll[1]) {
  player1Score += 1;
} else if (currentRoll[0] > currentRoll[1]) {
  player0Score += 1;
}
```
and remove the `else`.

script.js
```diff
@@ -1,4 +1,133 @@
...
+// roll dice
+btnRollDice.addEventListener('click', function () {
+  // let input = inputEl.value;
+  // outputEl.textContent = input;
+  rollDice();
+});
+
+// change player in the game
+const changePlayer = function () {
+  activePlayer = activePlayer === 0 ? 1 : 0;
```
leechuanxin, line 153: How comfortable are you with the concept of "[truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) vs [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) values", and converting Boolean values back into numbers? :P

Just for fun (and experiment's sake), I am suggesting an alternative to your ternary operator here:

```
activePlayer = Number(activePlayer === 0);
```

The above line works because if `(activePlayer === 0)` is `true`, it will be converted to `1` (`Number(true)` defaults to `1`). If not, it will be converted to `0`.

Feel free to ignore my suggestion for this though! Your version is definitely more readable (better); I was just musing over techniques to circumvent using a ternary operator here, because if you pseudo-code line 113 it is as good as saying
```
// activePlayer is 1 (truthy) if true, else activePlayer is 0 (falsy)
// which may also sound like this in English: if it's true, it's true, else it's false
```
which _may_ sound like it defeats the purpose of using a ternary operator here.

Again, just for fun though! Your version is what most, if not all, developers will use.

script.js
```diff
@@ -1,4 +1,133 @@
...
+  instructionsEl.textContent = `${nameOfPlayers[activePlayer]}, please roll dice to start the game! 🎲`;
+};
+
+// get a random number
+const randomNumberGenerator = function () {
+  return Math.trunc(Math.random() * 6 + 1);
```
leechuanxin, line 159: Very nice use of `Math.trunc`!

script.js
```diff
@@ -1,4 +1,133 @@
...
+};
+
+// determines what happens when buttons are pressed and to change player after 1 turn is over.
+const gameLogic = function () {
+  // show current score for active player
+  document.getElementById(`current-${activePlayer}`).textContent =
+    currentRoll[`${activePlayer}`];
```
leechuanxin: Could have left it as `currentRoll[activePlayer]`. No need to parse `activePlayer` as a template literal here.

leechuanxin (commented):
Hey Kenneth, great job with Base and playing around with the DOM for Comfortable! Nothing much to add besides minor stuff to clean up / refactor at the moment.

Understand that you may also be working on the More Comfortables later (should you desire to), so certain parts of my suggestions may not be applicable by then. 

There's also a "just for fun" extracurricular comment somewhere below about truthy and falsy values (pertaining to your use of ternary operators), but feel free to ignore that!

script.js
kennethongcs: Think I was creating an expression for draw, but forgot to add it in.

script.js
kennethongcs: Thanks for expounding on the truthy and falsy values in a way that I would not have used them.

PR #187, opened 2021-10-19
Richie Chia basics-11-7

richiechia:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

4

-

Integrating increasing number of players

Comfort Level (1-5):
4

Completeness Level (1-5):
3

Great. Alot of logical thinking

Able to solve it thus far

### DIFF
--- a/index.html
+++ b/index.html
@@ -51,7 +51,7 @@
   <body>
     <h1 id="header">Basics: Beat That! 🚀</h1>
     <div id="container">
-      <p>Input:</p>
+      <p>Input: (First select number of players)</p>
       <input id="input-field" />
       <br />
       <button id="submit-button">Submit</button>
--- a/script.js
+++ b/script.js
@@ -1,4 +1,93 @@
+var plays = 1; // Number of plays
+var internal_counter = 0; // Rotate back and forth first and second condition
+var player_dice = ""; // Variable used to store the 2 dice
+
+var number_of_players = 0; // Used to get the number of players
+var list_of_numbers = [];
+var winner = "";
+
 var main = function (input) {
-  var myOutputValue = 'hello world';
+  var myOutputValue = "";
+  // Key in the number of players as long as below 2. Cannot proceed
+  if (number_of_players < 2) {
+    console.log(!input);
+    if (!input) {
+      myOutputValue =
+        "Please fill number of players. Minimum 2! Don't leave it blank!";
+    } else if (Number.isNaN(Number(input))) {
+      myOutputValue =
+        "Please fill number of players. Minimum 2! Don't fill in letters!";
+    } else {
+      number_of_players = Number(input);
+      if (number_of_players < 2) {
+        myOutputValue = `The number of players is: ${number_of_players}. It is insufficient. Minimum 2!`;
+      } else {
+        myOutputValue = `The number of players is: ${number_of_players}. <br> Let's begin.<br><br> Player 1 starts. <br> Click 'Submit' to roll 2 dices.`;
+      }
+    }
+    return myOutputValue;
+  }
+
+  while (plays < number_of_players + 1) {
+    if (internal_counter % 2 == 0) {
+      var dice_holder = [];
+      var dice1 = diceRoll();
+      var dice2 = diceRoll();
+      // Append into the list
+      dice_holder.push(dice1, dice2);
+      player_dice = dice_holder;
+      console.log(`Player rolled: ${player_dice}`);
+
+      myOutputValue = `Welcome Player ${plays}. <br> You rolled:<br> Dice 1: ${dice1}. <br> Dice 2: ${dice2}. <br> Choose the order of the dice. <br><br> Choose '1' for dice 1 first<br> Choose '2' for dice 2 first `;
+
+      internal_counter += 1;
+      return myOutputValue;
+    } else {
+      if (Number(input) != 1 && Number(input) != 2) {
+        myOutputValue = `Please select '1' or '2' to choose dices 1 or 2 respectively.<br><br> You rolled:<br> Dice 1: ${player_dice[0]}.<br> Dice 2: ${player_dice[1]}.`;
+        return myOutputValue;
+      }
+
+      internal_counter += 1;
+      var player_select = Number(input);
+      if (player_select == 1) {
+        // if player selected first dice, extract out first dice from array.
+        first_digit = player_dice[player_select - 1];
+        second_digit = player_dice[player_select];
+      } else {
+        // if player selected second dice, extract out second dice from array
+        first_digit = player_dice[player_select - 1];
+        second_digit = player_dice[player_select - 2];
+      }
+      value = first_digit.toString() + second_digit.toString();
+      list_of_numbers.push(Number(value));
+      console.log(list_of_numbers);
+      // Let users know what is their actual value after selection
+      myOutputValue = `Player ${plays}, you chose Dice ${player_select} first. <br> Your number is: ${value}.`;
+      plays += 1;
+
+      // Adds the next player statement only if there is a next player.
+      if (plays <= number_of_players) {
+        myOutputValue += `<br> It is now Player ${plays}'s turn. <br><br> Click 'Submit' to roll 2 dices. ${winner}`;
+      }
+      if (plays == number_of_players + 1) {
+        location_of_index = list_of_numbers.indexOf(
+          Math.max(...list_of_numbers)
+        );
+        winner = `<br><br> The winner is Player ${location_of_index + 1}`;
+      }
+      myOutputValue += winner;
+      return myOutputValue;
+    }
+  }
+
+  myOutputValue = `The game is completed! ${winner}! Congratulations!`;
   return myOutputValue;
+};
+
+var diceRoll = function () {
+  randomDecimal = Math.random() * 6;
+  randomInteger = Math.floor(randomDecimal);
+  dice = randomInteger + 1;
+  return dice;
 };
--- a/script.js
+++ b/script.js
@@ -3,12 +3,15 @@ var internal_counter = 0; // Rotate back and forth first and second condition
 var player_dice = ""; // Variable used to store the 2 dice
 
 var number_of_players = 0; // Used to get the number of players
-var list_of_numbers = [];
+var list_of_scores = [];
 var winner = "";
+var round = 0;
+
+game_mode = 0;
 
 var main = function (input) {
   var myOutputValue = "";
-  // Key in the number of players as long as below 2. Cannot proceed
+  //// Key in the number of players as long as below 2. Cannot proceed /////
   if (number_of_players < 2) {
     console.log(!input);
     if (!input) {
@@ -22,12 +25,31 @@ var main = function (input) {
       if (number_of_players < 2) {
         myOutputValue = `The number of players is: ${number_of_players}. It is insufficient. Minimum 2!`;
       } else {
-        myOutputValue = `The number of players is: ${number_of_players}. <br> Let's begin.<br><br> Player 1 starts. <br> Click 'Submit' to roll 2 dices.`;
+        myOutputValue = `The number of players is: ${number_of_players}. <br> Let's begin.<br><br> Please fill in a game mode. <br><br> '1' - Highest Number Wins <br> '2' - Lowest Number Wins <br> `;
+        //Create the score_list
+        list_of_scores = new Array(number_of_players).fill(0);
+        console.log(`Generated a list of scores: ${list_of_scores}`);
       }
     }
     return myOutputValue;
   }
+  ////////////////////// Ask for game mode ///////////////////////
+  if (!game_mode) {
+    game_mode = Number(input);
+    myOutputValue = `You've selected Game Mode: ${game_mode}.<br><br> Press 'Submit' to roll 2 dices. `;
+    if (!input) {
+      myOutputValue =
+        "Please fill in a game mode. <br><br> '1' - Highest Number Wins <br> '2' - Lowest Number Wins <br>";
+    } else if (game_mode != 1 && game_mode != 2) {
+      myOutputValue =
+        "Please fill in either '1' or '2'. <br><br>'1' - Highest Number Wins <br> '2' - Lowest Number Wins <br>";
+      game_mode = 0;
+    }
+    return myOutputValue;
+  }
+  console.log("The game mode is: " + game_mode);
 
+  ////////////////////////// Game Play //////////////////////////////////
   while (plays < number_of_players + 1) {
     if (internal_counter % 2 == 0) {
       var dice_holder = [];
@@ -36,20 +58,30 @@ var main = function (input) {
       // Append into the list
       dice_holder.push(dice1, dice2);
       player_dice = dice_holder;
-      console.log(`Player rolled: ${player_dice}`);
+      console.log(`Player ${plays} rolled: ${player_dice}`);
 
-      myOutputValue = `Welcome Player ${plays}. <br> You rolled:<br> Dice 1: ${dice1}. <br> Dice 2: ${dice2}. <br> Choose the order of the dice. <br><br> Choose '1' for dice 1 first<br> Choose '2' for dice 2 first `;
+      myOutputValue = `Welcome Player ${plays}. <br> You rolled:<br> Dice 1: ${dice1}. <br> Dice 2: ${dice2}. <br> Choose the order of the dice. <br><br> Choose '1' for dice 1 first<br> Choose '2' for dice 2 first. <br><br> Otherwise, it'll be automatically optimise for you to win.`;
 
       internal_counter += 1;
       return myOutputValue;
     } else {
-      if (Number(input) != 1 && Number(input) != 2) {
+      // check the largest number or smallest number within the 2 dices. Depending on the game_mode
+      index_of_dice_win = Number(winning_gamemode(player_dice, game_mode));
+      console.log(player_dice[index_of_dice_win]);
+
+      var player_select = index_of_dice_win + 1;
+      // The 5 lines below maybe redundant.
+      if (player_select != 1 && player_select != 2) {
         myOutputValue = `Please select '1' or '2' to choose dices 1 or 2 respectively.<br><br> You rolled:<br> Dice 1: ${player_dice[0]}.<br> Dice 2: ${player_dice[1]}.`;
         return myOutputValue;
-      }
-
+      } // redundant
       internal_counter += 1;
-      var player_select = Number(input);
+
+      // Allows manual overwrite instead of automatically select
+      if (Number(input) == 1 || Number(input) == 2) {
+        player_select = Number(input);
+      }
+      // var player_select = Number(input);
       if (player_select == 1) {
         // if player selected first dice, extract out first dice from array.
         first_digit = player_dice[player_select - 1];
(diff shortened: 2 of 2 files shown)


index.html
```diff
@@ -51,7 +51,7 @@
   <body>
     <h1 id="header">Basics: Beat That! 🚀</h1>
     <div id="container">
-      <p>Input:</p>
+      <p>Input: (First select number of players)</p>
```
leechuanxin, line 54: Nice tampering with the HTML!

1 suggestion, I will add "First select number of players" to `output-div` in [line 59](https://github.com/rocketacademy/basics-beat-that/blob/aa261cb057ab351f5fd99b9da1068b41dda55703/index.html#L59) instead.

Ultimately, the `main` function will print its `return` value into `output-div` through the lifecycle of the program. This is on a "need-to-know" basis now, until the end of the course. If you are curious to know more, you can have a look at [lines 65 to 77](https://github.com/rocketacademy/basics-beat-that/blob/aa261cb057ab351f5fd99b9da1068b41dda55703/index.html#L59) and try to understand them, or read ahead [here](https://basics.rocketacademy.co/12-next-steps/12.1-browser-applications-with-dom). :)

My suggestion of moving your sentence into `output-div` means the "First select number of players" text will only appear at the start of the program (on page load). It will be the default string in `output-div` when the page loads, up until you start clicking Submit.

My suggestion will still achieve what I think you have intended for this to achieve, except that you won't have the "Input: (First select number of players)" text  continually dangling in the `<p>` even after changing game states to one where you no longer select number of players.

script.js
```diff
@@ -1,4 +1,291 @@
...
+  score_msg = generate_final_scorecard(list_of_scores);
+  myOutputValue = `Round ${round} is completed! This is game mode ${game_mode}. <br><br> <b>Leaderboard</b> <br> ${score_msg} ${winner} <br><br> Click 'Submit' to continue onto next round`;
+  console.log(myOutputValue);
+  winner = "";
+
   return myOutputValue;
 };
+
+var check_input_players = function (number_of_players, input) {
+  state = false;
```
leechuanxin, line 149: Despite the convenience of creating variables without using the `var` keywords (as in [line 149](https://github.com/rocketacademy/basics-beat-that/blob/aa261cb057ab351f5fd99b9da1068b41dda55703/script.js#L149)), I strongly advise against it, especially when working on large programs. Python doesn't really separate "re-assignment" and "creation" in syntax, but JavaScript has very weird behaviours if the 2 are not performed properly.

Creating variables within a function scope within `var` will cause JavaScript to assume you are creating it within the global scope instead. This is assuming the next "upper scope" is the global scope, and there's no such identically-named variable already created in the global scope. Otherwise it will be treated as a re-assignment of values in the next "upper scope". 

What then, you might ask, about global variables? Since JavaScript assumes variables created not using `var` is in the global scope, won't it be convenient to just to create global variables without `var`?

As mentioned in your in-class exercise on Day 6, it makes code hard to read, especially in developer handoffs. If I want to monitor the state (of changes) for a particular global variable, I may want to start at where it is first created. Most developers will probably start with "Cmd/Ctrl+F"-ing your code for `var someGlobalVariable ...`.

Moreover, things may not seem that bad now with a single `script.js` file in your `index.html`. However, in complex web applications, it is likely you will see multiple `.js` files in a single HTML file. Let's assume there's another `<script>` tag before [line 62 in your index.html](https://github.com/richiechia/basics-beat-that/blob/main/index.html#L62), which points to another `.js` file, say a `script_before.js`. There may be another global variable already created with the same variable name in that file. Then, JavaScript will just assume you are attempting to re-assign values to the identically-named global variable in `script_before.js`.

For more reading about whether it's "optional" to use `var`, you can view the short comment here: [https://stackoverflow.com/a/2485439](https://stackoverflow.com/a/2485439). If you have the chance, the other comments on the page also explain the complex scoping rules (when JavaScript interprets it as an assignment at an upper scope vs when it just assumes you are creating it in the global scope) in a more succinct manner than I am doing now.

script.js
```diff
@@ -1,4 +1,291 @@
...
+  if (number_of_players < 2) {
```
leechuanxin, line 150: I noticed that you have the `number_of_players < 2` condition double-checked here, and [line 160](https://github.com/rocketacademy/basics-beat-that/blob/aa261cb057ab351f5fd99b9da1068b41dda55703/script.js#L160).

I played around with refactoring it so that you have less nested `if` blocks, and only check `if (number_of_players < 2)` once.

Consider this for starters (might have to change the other parts of your code too, because I realised the program breaks when I refactored it this way):

```
var check_input_players = function (input) {
  var state = false;
  var myOutputValue = "Please fill number of players. Minimum 2! ";
  var newNumberOfPlayers = Number(input);

  // input is empty, or just a series of spaces
  if (input.trim() == "") {
    myOutputValue += "Don't leave it blank!";
  }
  // input is not a number
  else if (Number.isNaN(newNumberOfPlayers)) {
    myOutputValue += "Don't fill in letters!";
  }
  // we know input here should be a number, test if it's less than 2
  else if (newNumberOfPlayers < 2) {
    myOutputValue = `The number of players is: ${newNumberOfPlayers}. It is insufficient. Minimum 2!`;
  }
  // all other cases, input should be a number of 2 or more now
  else {
    myOutputValue = `The number of players is: ${newNumberOfPlayers}. <br> Let's begin.<br><br> Please fill in a game mode. <br><br> '1' - Highest Number Wins <br> '2' - Lowest Number Wins <br> `;
    //Create the score_list based on the number of players. Fill it with '0's
    list_of_scores = new Array(Number(newNumberOfPlayers)).fill(0);
    console.log(`Generated a list of scores: ${list_of_scores}`);
    state = true;
  }
  return [myOutputValue, state, newNumberOfPlayers];
};
```

script.js
```diff
@@ -1,4 +1,291 @@
...
+    console.log(!input);
+    if (!input) {
+      myOutputValue =
+        "Please fill number of players. Minimum 2! Don't leave it blank!";
+    } else if (Number.isNaN(Number(input))) {
+      myOutputValue =
+        "Please fill number of players. Minimum 2! Don't fill in letters!";
+    } else {
+      number_of_players = Number(input);
```
leechuanxin, line 159: `number_of_players` is a parameter to your `check_input_players` function. Typically, we don't recommend re-assigning values to parameters. ESLint explains why this is bad practice [here](https://eslint.org/docs/rules/no-param-reassign).

What I suggest doing:
```
var check_input_players = function (number_of_players, input) {
  var newNumberOfPlayers = number_of_players;
  ...
  newNumberOfPlayers = Number(input);
  ...
  return [myOutputValue, state, newNumberOfPlayers];
}
```

Besides, `number_of_players` will be storing its new value (from the input) in [line 29](https://github.com/rocketacademy/basics-beat-that/blob/aa261cb057ab351f5fd99b9da1068b41dda55703/script.js#L29) anyway. There's no need to re-assign the parameter from within your check_input_players function.

script.js
```diff
@@ -1,4 +1,291 @@
...
+    }
+    check_done_input = check_input[1];
+    myOutputValue =
+      "Please fill in a game mode. <br><br> '1' - Highest Number Wins <br> '2' - Lowest Number Wins <br>";
+    number_of_players = check_input[2];
+    return myOutputValue;
+  }
+
+  ////////////////////// Ask for game mode ///////////////////////
+  if (!game_mode) {
```
leechuanxin, line 34: The 'ask for game mode' game state can be reduced to a function, since the final outcome is just printing the results of the user's input (in terms of whether a valid game mode is chosen).

Consider the following lines in the `if` block:
```
if (!game_mode) {
  return selectGameMode(input);
}
```
and the declaration of the `selectGameMode` function:
```
var selectGameMode = function (input) {
  // create the "number" version of input for checks,
  // but don't assign it to game_mode yet
  var numberInput = Number(input);
  var output = `You've selected Game Mode: ${numberInput}.<br><br> Press 'Submit' to choose the number of dices. `;
  var instructionsString =
    " <br><br> '1' - Highest Number Wins <br> '2' - Lowest Number Wins <br>";
  if (input.trim() == "" || Number.isNaN(numberInput)) {
    output = "Please fill in a game mode." + instructionsString;
  } else if (numberInput != 1 && numberInput != 2) {
    output = "Please fill in either '1' or '2'." + instructionsString;
  } else {
    // set global game_mode only when you have verified the input
    game_mode = numberInput;
  }
  return output;
};
```

script.js
```diff
@@ -1,4 +1,291 @@
...
+    game_mode = Number(input);
```
leechuanxin, line 35: I probably won't want to re-assign the global `game_mode`, if the `input` is not properly verified as valid yet.

Instead, you can store it in a placeholder variable first, such as `var numberInput = Number(input);`

This allows you to avoid having to re-assign `game_mode` back to its "safe" default value, as you are doing in [line 43](https://github.com/rocketacademy/basics-beat-that/blob/aa261cb057ab351f5fd99b9da1068b41dda55703/script.js#L43). Without assigning an invalid number directly into `game_mode` first, you can always know it stays "safe", ie. you know it should only be `0`, `1` or `2`.

This also means you will need an `else` in your conditional chain. When it's finally clear that the input is valid, assign the global `game_mode` to `numberInput`.

Consider the following:

```
if (!game_mode) {
    var numberInput = Number(input);
    ...
    if (!input) {
      ...
    } else if (game_mode != 1 && game_mode != 2) {
      ...
    } else {
      // set global game_mode only when you have verified the input
      game_mode = numberInput;
    }
    ...
  }
```

script.js
```diff
@@ -1,4 +1,291 @@
...
+    myOutputValue = `You've selected Game Mode: ${game_mode}.<br><br> Press 'Submit' to choose the number of dices. `;
+    if (!input) {
+      myOutputValue =
+        "Please fill in a game mode. <br><br> '1' - Highest Number Wins <br> '2' - Lowest Number Wins <br>";
+    } else if (game_mode != 1 && game_mode != 2) {
+      myOutputValue =
+        "Please fill in either '1' or '2'. <br><br>'1' - Highest Number Wins <br> '2' - Lowest Number Wins <br>";
+      game_mode = 0;
```
leechuanxin, line 43: May be redundant if you do not set `game_mode` directly in [line 35](https://github.com/rocketacademy/basics-beat-that/blob/aa261cb057ab351f5fd99b9da1068b41dda55703/script.js#L35).

script.js
```diff
@@ -1,4 +1,291 @@
...
+    myOutputValue = `You've selected ${no_of_dice} dices.<br><br> Press 'Submit' to roll ${no_of_dice} dices.`;
+    if (!input) {
+      myOutputValue = "Please fill in number of dices. From 1 to infinity.";
+    } else if (no_of_dice < 1) {
+      myOutputValue = `Please input a positive value from 1 to infinity.`;
+      no_of_dice = 0;
+    }
+    console.log("The number of dice is: " + no_of_dice);
+    return myOutputValue;
+  }
```
leechuanxin, line 61: Same feedback as my comments about the code in the `if (!game_mode)` block.

Consider this in the new `if` block:

```
if (!no_of_dice) {
    return selectNumberOfDices(input);
  }
```

and this for the new `selectNumberOfDices` function:

```
var selectNumberOfDices = function (input) {
  var numberInput = Number(input);
  var output = `You've selected ${numberInput} dices.<br><br> Press 'Submit' to roll ${numberInput} dices.`;
  if (input.trim() == "" || Number.isNaN(numberInput)) {
    output = "Please fill in number of dices. From 1 to infinity.";
  } else if (numberInput < 1) {
    output = `Please input a positive value from 1 to infinity.`;
  } else {
    no_of_dice = numberInput;
  }
  return output;
};
```

script.js
```diff
@@ -1,4 +1,291 @@
...
+      // Update the scores in the array
+      list_of_scores[plays - 1] += Number(value);
+      console.log(list_of_scores);
+      // Let users know what is their actual value after selection
+      myOutputValue = `Player ${plays}<br> Your number is: ${value}.`;
+      plays += 1;
+
+      // Adds the next player statement only if there is a next player.
+      if (plays <= number_of_players && game_mode_knockout == false) {
+        myOutputValue += `<br> It is now Player ${plays}'s turn. <br><br> Click 'Submit' to roll 2 dices. ${winner}`;
```
leechuanxin, line 95: Just a minor display error in your copywriting here. "2" is currently hard-coded as a value, I think what you are looking for is:

```
myOutputValue += `<br> It is now Player ${plays}'s turn. <br><br> Click 'Submit' to roll ${no_of_dice} dices. ${winner}`;
```

script.js
```diff
@@ -1,4 +1,291 @@
...
+      myOutputValue =
+        "Please fill number of players. Minimum 2! Don't fill in letters!";
+    } else {
+      number_of_players = Number(input);
+      if (number_of_players < 2) {
+        myOutputValue = `The number of players is: ${number_of_players}. It is insufficient. Minimum 2!`;
+      } else {
+        myOutputValue = `The number of players is: ${number_of_players}. <br> Let's begin.<br><br> Please fill in a game mode. <br><br> '1' - Highest Number Wins <br> '2' - Lowest Number Wins <br> `;
+        //Create the score_list based on the number of players. Fill it with '0's
+        list_of_scores = new Array(number_of_players).fill(0);
```
leechuanxin, line 165: Very very nice use of the `Array.fill` method to initialise score values based on the number of players.

script.js
```diff
@@ -1,4 +1,291 @@
...
+  ////////////////////////// Game Play //////////////////////////////////
+  while (plays < number_of_players + 1) {
+    if (internal_counter % 2 == 0) {
+      var dice_holder = [];
+      var dice_msg = "";
+      // Fix the dice at 2 first
+      player_dice = generate_no_of_dice(dice_holder, no_of_dice);
+      console.log(`Player ${plays} rolled: ${player_dice}`);
+
+      a = 0;
```
leechuanxin, line 72: Create a variable here using `var`.

script.js
```diff
@@ -1,4 +1,291 @@
...
+      while (a < no_of_dice) {
+        dice_msg += `<br> Dice ${a + 1}: ${player_dice[a]}.`;
+        a++;
```
leechuanxin, line 75: Suggest using `a += 1` instead of `a++`. Using the `++` syntax perfectly seems very dependent on the lack of human error in writing your code, and ESLint [describes very well what happens if you make even the slightest syntactical mistake when using it](https://eslint.org/docs/rules/no-plusplus).

`a += 1` is a lot safer to use in general, and we recommend using it for your `for` loops too (if you intend to use those).

```
for (var i = 0; i < 5; i += 1) {
  ...
}
```

script.js
```diff
@@ -1,4 +1,291 @@
...
+      }
+
+      myOutputValue = `Welcome Player ${plays}. ${dice_msg} <br><br> Dice will be automatically optimise for you to win. Click 'Submit' to continue to see your value`;
+
+      internal_counter += 1;
+      return myOutputValue;
+    } else {
+      internal_counter += 1;
+      // Sort and obtain the final output value
+      value = optimise_values_of_dice(player_dice, game_mode);
```
leechuanxin, line 85: Create a variable here using `var`.

script.js
```diff
@@ -1,4 +1,291 @@
...
+      myOutputValue = `Please input a positive value from 1 to infinity.`;
+      no_of_dice = 0;
+    }
+    console.log("The number of dice is: " + no_of_dice);
+    return myOutputValue;
+  }
+
+  ////////////////////////// Game Play //////////////////////////////////
+  while (plays < number_of_players + 1) {
+    if (internal_counter % 2 == 0) {
```
leechuanxin, line 65: Interesting use of an `internal_counter` to oscillate between 'what Player X has rolled' and 'what is the optimised combined number from Player X's rolls'.

I will also suggest modularising the logic of the aforementioned "sub-modes" into functions.

script.js
```diff
@@ -1,4 +1,291 @@
...
+        }
+        myOutputValue += winner;
+      }
+
+      return myOutputValue;
+    }
+  }
+  // Reset the game
+  internal_counter = 0;
+  plays = 1;
```
leechuanxin, line 137: May want to consider dumping all of your variable re-assignments for "resetting the game" in a function. Just so you can always re-use a `resetGame()` instead of re-writing lines 136 and 137 every time you want to do a reset.

script.js
```diff
@@ -1,4 +1,291 @@
...
+  }
+  return msg;
+};
+
+var max_value_repetition = function (list_of_scores) {
+  j = 0;
+  count = 0;
+  var msg = "";
+
+  largest_value = Math.max(...list_of_scores);
```
leechuanxin, line 200: Nice use of `Math.max` and the spread syntax.

script.js
```diff
@@ -1,4 +1,291 @@
...
+    dice_holder.push(diceRoll());
+    k++;
+  }
+  return dice_holder;
+};
+
+// Obtain the correct value either highest or lowest
+var optimise_values_of_dice = function (player_dice, game_mode) {
+  // sort the values from smallest to largest
+  var numArray = new Float64Array(player_dice);
```
leechuanxin, line 244: Curious as to why you would need to use a `Float64Array` as opposed to a regular `Array` for this.

script.js
```diff
@@ -1,4 +1,291 @@
...
+  numArray = numArray.sort();
```
leechuanxin, line 245: Oh right, I see why you are using `Float64Array` now. So that you can sort it numerically instead of the default `Array.sort`'s lexicographical manner.

Most developers will do it this way:

```
// spread syntax to make a shallow copy of existing player_dice
var numArray = [...player_dice];
// no need to re-assign numArray here, .sort sorts the existing array in place
// this should apply even for sorting Float64Arrays
numArray.sort(function(a, b) {
  return a - b;
})
```

Your way seems fine, you can stick to that. Be sure to add comments explaining why you set it up this way though, ie. to sort numerically as opposed to the default lexicographical manner. Most developers will probably not touch `Float64Array`s throughout their lifetime, so explaining this to them will help.

script.js
```diff
@@ -1,4 +1,291 @@
...
+    return true;
+  }
+  return false;
+};
+
+var winning_gamemode = function (list_of_scores, game_mode) {
+  winning_condn = 0;
+
+  if (game_mode == 1) {
+    winning_condn = list_of_scores.indexOf(Math.max(...list_of_scores));
```
leechuanxin, line 220: Nice use of `Array.indexOf` to find the index of the lowest / highest numbers.

leechuanxin (commented):
Thanks for setting up your PR early! I appreciate you going far into experimenting this Project, including the use of Array prototype methods that are out of the curriculum. I learnt some cool things from your set up too; I had to review my notes on `Float64Array` just to be doubly clear why you were using it vs. regular Arrays.

Besides my specific comments below (which may take quite a bit of refactoring to resolve), I would like you to stick to certain conventions for naming your variables. Ordinary variables in JavaScript should be named in [lowerCamelCase](https://basics.rocketacademy.co/3-basic-data-manipulation/3.2-variables#variable-naming-convention). Right now, I am seeing a mix of lowerCamelCases, and lower_snake_cases.

Noticeably, I have been very harsh on you about adopting best practices for this Project. The reasons are:
1.  It's easier to fix bad practices early in one's learning journey vs. much later as a more senior developer.
2.  I believe the quality of one's learning journey is built upon strong fundamentals. It may be tricky to pick up advanced topics if you are continually side-tracked by having to resolve bad practices or gaps in knowledge later in your learning journey.
3. It's tricky in a twice-a-week 2-hour class to learn best practices together with new syntax, especially when "experience is the best teacher" for the topic. In-classes aren't sufficient besides practising the new concepts (usually syntax) you just learnt, and projects are the only time I can properly have a look at your conduct in structuring heavier logic. 
4. You don't seem to have a problem with learning prototype methods or picking up new syntax on your own. I think you are capable to begin thinking about the 'why' as opposed to just the 'how to' in implementation :). A lot of what I am saying here is probably out of the Basics curriculum, but it's never too early to begin thinking about them.

Have a look at these for starters, and I am looking forward to seeing improvements in your code quality for subsequent exercises and your final Project! Feel free to contact me if you have any questions about any comment here.

PR #201, opened 2021-10-26
Wayne Wee basics-11-7

Waynerbee:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

How many hours did you spend on this assignment? 1

What part of the assignment did you spend the most time on? Getting the states right 

Comfort Level (1-5): 4

Completeness Level (1-5): 5

What did you think of this deliverable? It was ok for me, but I think some of my peers felt that it might be easier if its broken up into more phrases.

Is there anything in this code that you feel pleased about? I am glad it works! haha

### DIFF
--- a/script.js
+++ b/script.js
@@ -57,7 +57,7 @@ var playerSelects = function (diceNumber) {
   console.log(state);
   var output = "Please reenter Dice '1' or '2'";
   if (diceNumber == "1") {
-    combinedNumber = String(diceOne) + String(diceTwo);
+    var combinedNumber = String(diceOne) + String(diceTwo);
     output =
       "Player " +
       playerNumber +
@@ -71,7 +71,7 @@ var playerSelects = function (diceNumber) {
     numbers.push(Number(combinedNumber));
     state = "state3";
   } else if (diceNumber == "2") {
-    combinedNumber = String(diceTwo) + String(diceOne);
+    var combinedNumber = String(diceTwo) + String(diceOne);
     output =
       "Player " +
       playerNumber +
@@ -105,7 +105,7 @@ var playerOption = function (condition) {
     diceOne = generateRandomDiceInteger();
     diceTwo = generateRandomDiceInteger();
 
-    var output =
+    output =
       "Welcome player " +
       playerNumber +
       ".<br> You rolled " +
@@ -127,7 +127,7 @@ var playerOption = function (condition) {
     });
 
     var position = numbers.indexOf(highestNumber) + 1;
-    var output =
+    output =
       "The highest number is " +
       highestNumber +
       ". Player " +
@@ -142,7 +142,7 @@ var playerOption = function (condition) {
     diceOne = 0;
     diceTwo = 0;
     playerNumber = 1;
-    score = [];
+    numbers = [];
   }
 
   return output;
--- a/script.js
+++ b/script.js
@@ -1,4 +1,164 @@
+// global variables
+
+var state = "state1";
+var diceOne = 0;
+var diceTwo = 0;
+var playerNumber = 1;
+var numbers = [];
+
+/*
+
+state1: Player rolls (Input: (anything))
+state2: Player picks which is the first dice (Input: "first") and which is the second (Input: "second")
+state3: Proceed to results (Input: "show results") or add a player (Input: "next player")
+state4: Results is shown and reset initial variables
+
+*/
+
+// random dice number
+
+var generateRandomDiceInteger = function () {
+  var randomDec = Math.random() * 6;
+  var randomNum = Math.ceil(randomDec);
+  return randomNum;
+};
+
+// state1
+
+var playerRolls = function () {
+  // reset variables
+  console.log(state);
+  diceOne = 0;
+  diceTwo = 0;
+
+  diceOne = generateRandomDiceInteger();
+  diceTwo = generateRandomDiceInteger();
+  state = "state2";
+
+  console.log(diceOne);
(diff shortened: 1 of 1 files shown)


script.js
```diff
@@ -1,4 +1,164 @@
...
+    output =
+      "Player " +
+      playerNumber +
+      ", you chose Dice " +
+      diceNumber +
+      " first.<br>" +
+      "Your number is " +
+      combinedNumber +
+      ".<br>" +
+      "Enter 'add player' to add new players or 'results' to show results.";
```
leechuanxin, line 70: Not really a technical issue, but I would change the copy here to say `... 'show result' to show results ...`. When testing your game, I kept typing in 'results' accidentally and clicking Submit, only for your interface to tell me I should be typing `'show result'` instead 😅

script.js
```diff
@@ -1,4 +1,164 @@
...
+state3: Proceed to results (Input: "show results") or add a player (Input: "next player")
+state4: Results is shown and reset initial variables
+
+*/
+
+// random dice number
+
+var generateRandomDiceInteger = function () {
+  var randomDec = Math.random() * 6;
+  var randomNum = Math.ceil(randomDec);
```
leechuanxin, line 22: `Math.random()` will generate a number from `0` to `0.9999...`. Multiplied by 6, it gives you a range of `0` to `5.999...`.

I will be careful about using `Math.ceil` here. This is because of the _very very rare case_ where your `randomDec` value is still `0` after multiplying it by 6.

`Math.ceil(0.0000...1)` gives `1`, but `Math.ceil(0)` still gives `0`. This means your `randomNum` will be in a range of `0 to 6` as opposed to the `1 to 6` range expected from dice rolls.

I would still keep to the original random dice roll algorithm, like so:

```
var rollDice = function () {
  // produce a decimal between 0 (inclusive) and 6 (exclusive)
  var randomDec = Math.random() * 6;
  // remove the decimal
  var randomInt = Math.floor(randomDecimal);
  // add 1 to get a number between 1 and 6 inclusive
  var randomNum = randomInteger + 1;
  return randomNum;
};
```

script.js
```diff
@@ -1,4 +1,164 @@
...
+  return output;
+};
+
+// main fucntion
 var main = function (input) {
-  var myOutputValue = 'hello world';
+  var myOutputValue = "There is an error in the states.";
+
+  if (state == "state1") {
+    myOutputValue = playerRolls();
```
leechuanxin, line 156: Great to see you modularising your states into functions! It really makes the `main` function look clean.

I will probably just name the strings in the states more clearly, like `"player rolls"` or `"select number order"`, instead of `"state1"`, `"state2"` ... etc. It makes developer handoffs much easier, so other developers working on your code won't have to second-guess what `"state1"`, `"state2"` etc actually refer to.

script.js
```diff
@@ -1,4 +1,164 @@
+// global variables
+
+var state = "state1";
+var diceOne = 0;
+var diceTwo = 0;
```
leechuanxin, line 5: Since we have already covered arrays in the previous few lessons, why not try storing dice rolls in a single array, like `var dices = [0, 0];`? 

It may make management of your dice rolls a lot easier if you intend to progress past Base, especially if you intend to try the [More Comfortable requirement of Variable Number of Dices](https://basics.rocketacademy.co/projects/project-2-beat-that#variable-number-of-dice).

script.js
```diff
@@ -1,4 +1,164 @@
...
+var playerNumber = 1;
```
leechuanxin, line 6: Good use of a number to store the current player turn!

Most students tend to make their jobs a little more difficult by storing strings instead, such as `var playerNumber = "Player 1";`. Numbers are definitely easier to manage for storing player turns, especially if you intend to go into heavier requirements such as [Variable Number of Players](https://basics.rocketacademy.co/projects/project-2-beat-that#variable-number-of-dice) (especially if your interface requires the players to say the number of players they want right from the beginning, as opposed to after every turn like you did)

script.js
```diff
@@ -1,4 +1,164 @@
...
+      "Enter anything to continue.";
+    console.log("Position:" + position);
+    console.log("Position:" + highestNumber);
+
+    // reset initial variables
+    state = "state1";
+    diceOne = 0;
+    diceTwo = 0;
+    playerNumber = 1;
+    numbers = [];
```
leechuanxin, line 145: Nicely done on being able to identify variables you need to reset moving on to new rounds / games.

Only a small suggestion, to have this reset logic in a function itself. Currently, the way you have set up your interface doesn't require it to be a function, since the player decides if they want to continue adding players only after the player's turn ends.

However, should you change your interface such that the player(s) decide from the very start of the game how many players will be playing, you may have to do resets a few times and modularising this part into a function will help.

leechuanxin (commented):
Hey Wayne,

Overall your logic seems pretty okay. Main feedback will just be the way you set up your interface, especially the consistency of telling the players to type in "results" or "show result" to show the result. 😅

It will be great if the strings related to the different game states are named properly too, just for developer handoff purposes. It is not required in the context of Basics because projects are solo work, but it is just a good practice to get into.

Lastly, be careful of your `generateRandomDiceInteger` function logic. You are using `Math.ceil`, but you may be forgetting the _very rare case_ where `Math.random` still gives you a flat whole number `0`. Sure, such cases are rare, but it's good practice to always account for such rare cases (we call these edge cases) so that it doesn't affect the game experience. Never trust luck :)

PR #206, opened 2021-10-27
Project 2 Beat That! 11-3

jason1015:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

6 hours

Math.max return NaN result

While loops

Comfort Level (1-5):
3

Completeness Level (1-5):
2

More examples of application will help

NA

### DIFF
--- a/script.js
+++ b/script.js
@@ -24,8 +24,11 @@ var auto = `Auto Standard Beat That! Game`;
 var autoReverse = `Auto Reverse Beat That! Game`;
 var variableDiceStandard = `Variable Dice, Auto Standard Beat That! Game`;
 var variableDiceReverse = `Variable Dice, Auto Reverse Beat That! Game`;
+var variablePlayerStandard = `Variable Player, Auto Standard Beat That! Game`;
+var variablePlayerReverse = `Variable Player, Auto Reverse Beat That! Game`;
 var currentGame = ``;
 var numOfDice = 0;
+var numOfPlayer = 1;
 
 // Initial game prompt
 {
@@ -38,7 +41,7 @@ var numOfDice = 0;
 // generate random dice number from 1 to 6
 var generateRandomDiceNumber = function () {
   var randomDiceNumber = Math.floor(Math.random() * 6) + 1;
-  console.log(`randomDiceNumber =${randomDiceNumber}`);
+
   return randomDiceNumber;
 };
 
@@ -183,6 +186,12 @@ var checkGameSelected = function () {
   if (gameMode == 6) {
     currentGame = variableDiceReverse;
   }
+  if (gameMode == 7) {
+    currentGame = variablePlayerStandard;
+  }
+  if (gameMode == 8) {
+    currentGame = variablePlayerReverse;
+  }
   return currentGame;
 };
 
@@ -231,11 +240,13 @@ var gameReset = function () {
   return `Game reset, please choose game mode to start<br>${gameInstruction1}`;
 };
 // game instruction message
-var gameInstruction1 = `Enter '1' for ${standard}<br>Enter '2' for ${reverse} Game<br>Enter '3' for ${auto} Game<br>Enter '4' for ${autoReverse} Game<br>Enter '5' for ${variableDiceStandard} Game<br>Enter '6' for ${variableDiceReverse} Game`;
+var gameInstruction1 = `Enter '1' for ${standard}<br>Enter '2' for ${reverse} Game<br>Enter '3' for ${auto} Game<br>Enter '4' for ${autoReverse} Game<br>Enter '5' for ${variableDiceStandard} Game<br>Enter '6' for ${variableDiceReverse} Game<br>Enter '7' for ${variablePlayerStandard} Game<br>Enter '8' for ${variablePlayerReverse} Game`;
 
 var gameInstruction2 = `Please enter the number '1' or '2'.`;
 
-var gameInstruction3 = `Please enter the number of dice`;
+var gameInstruction3 = `Please enter the number(>0) of dice`;
+
+var gameInstruction4 = `Please enter the number(>1) of players`;
 
 var main = function (input) {
   var myOutputValue = ``;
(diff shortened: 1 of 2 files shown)


script.js
```diff
@@ -1,4 +1,542 @@
...
+var reverse = `Reverse Beat That! Game`;
+var auto = `Auto Standard Beat That! Game`;
+var autoReverse = `Auto Reverse Beat That! Game`;
+var variableDiceStandard = `Variable Dice, Auto Standard Beat That! Game`;
+var variableDiceReverse = `Variable Dice, Auto Reverse Beat That! Game`;
+var variablePlayerStandard = `Variable Player, Auto Standard Beat That! Game`;
+var variablePlayerReverse = `Variable Player, Auto Reverse Beat That! Game`;
+var currentGame = ``;
+var numOfDice = 0;
+var numOfPlayer = 1;
```
JustinWong98, line 31: While global variables are helpful, I would suggest not having so many if possible - it can be hard to keep track of which variables hold what value with that many globals to keep in mind. While some are unavoidable like gamemodes and player scores, I think having the player dices stored as globals is a bit unnecessary - you could keep each player dice in an array, or even better keep both player dices as an array inside another array

script.js
```diff
@@ -1,4 +1,542 @@
...
+
+// Initial game prompt
+{
+  var color = "green";
+  document.write(
+    `<CENTER><FONT FACE=ARIAdL,VERDANA COLOR=${color} SIZE=5>Welcome!<br>Click Submit to Start Playing.</FONT><HR NOSHADE WIDTH=450></CENTER><P>`
+  );
```
JustinWong98, line 38: Impressive!

script.js
```diff
@@ -1,4 +1,542 @@
...
+  if (currentPlayer == player2 && playerOrderChoice == 1) {
+    player2Num = Number("" + player2Dice1 + player2Dice2);
+    console.log(player2Num);
+    return player2Num;
+  } else if (currentPlayer == player2 && playerOrderChoice == 2) {
+    player2Num = Number("" + player2Dice2 + player2Dice1);
+    console.log(player2Num);
+    return player2Num;
+  }
+};
```
JustinWong98, line 82: Interesting way of concatenating the numbers!

script.js
```diff
@@ -1,4 +1,542 @@
...
+  if (gameMode == 6) {
+    currentGame = variableDiceReverse;
+  }
+  if (gameMode == 7) {
+    currentGame = variablePlayerStandard;
+  }
+  if (gameMode == 8) {
+    currentGame = variablePlayerReverse;
+  }
+  return currentGame;
```
JustinWong98, line 195: for cases where you have a lot of if/else ifs comparing the same thing, you can use the switch statement:
https://www.w3schools.com/js/js_switch.asp

script.js
```diff
@@ -1,4 +1,542 @@
...
+    else if (
+      input != 1 &&
+      input != 2 &&
+      input != 3 &&
+      input != 4 &&
+      input != 5 &&
+      input != 6 &&
+      input != 7 &&
+      input != 8
+    ) {
```
JustinWong98, line 270: You can use ranges instead of having to list it out one by one:

`else if (input !>= 1 && input !<=8)`

script.js
```diff
@@ -1,4 +1,542 @@
...
+      // set gamemode
+      gameMode = Number(input);
+      console.log(gameMode);
+      return `${checkGameSelected()} is selected, enter number of dice to start playing`;
+    } else if (input == 7 || input == 8) {
+      // set gamemode
+      gameMode = Number(input);
+      console.log(gameMode);
+      return `${checkGameSelected()} is selected, enter number of player to start playing`;
+    }
```
JustinWong98, line 287: Is there a functional reason why you needed to separate out the inputs like this since the code is roughly the same? You could pass in the gameMode as a parameter to checkGameSelected() function and evaluate that in the helperfunction to decide what gamemode it is. That way, you don't even need this if else block

script.js
```diff
@@ -1,4 +1,542 @@
...
+      if (Number.isNaN(Number(input)) == true || !input || input == 1) {
+        // check if input is a number
+        return gameInstruction4;
+      } else {
+        numOfPlayer = input;
+        // switch game stage
+        gameStage = 2;
+        return `There will be ${numOfPlayer} players, click submit to start playing`;
+      }
+    }
```
JustinWong98, line 490: Likewise, it can be quite hard for someone to come in and decipher all this code since there are a lot of if and else ifs and nested ifs and else ifs. I think that naming the gameModes numerically can make it super difficult to decipher at first since they will need to keep referring to the list of gameModes while traversing through this code. I am not doubting that this code works but I think that many of the ifs and else ifs can be cut out

Also, we should keep the main function as short as possible - I think its best to segment out the game logic into its own helper functions. That way, people can see the helper function being called and see which gamemode they are currently reading about. It also helps to abstract the game logic away

script.js
```diff
@@ -1,4 +1,542 @@
...
+        console.log(`auto ${player2Num}`);
+        return player2Num;
+      } else if (player2Dice1 < player2Dice2) {
+        player2Num = Number("" + player2Dice1 + player2Dice2);
+        console.log(`auto ${player2Num}`);
+        return player2Num;
+      }
+    }
+  }
+};
```
JustinWong98, line 137: Since you are returning the number generated at the end of every if or else if statement (meaning that the code isn't reliant on any of the other code), and they are mostly similar, you can look into simply combining some of the statements. Also, it seems that the gameMode doesn't affect the order of the dice? I.E the code across the game modes seems the same.

JustinWong98 (commented):
Good effort! I understand that it is tough balancing so many game modes and implementing it all in one script, so I respect the effort you put in. I think a good step to take is that you could attempt to separate all game logic into its own helper functions instead of putting them in main. That way it could help you to better focus and have a clear idea on what you need to do for that particular game mode

PR #219, opened 2021-10-28
Batch 11 / Section 2: Ray H.

R4Y815:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

5-6 days? 40 hours?

1) Loop for returning combined numbers stop after first player, and output keeps showing same result of first combined number for other players under multiplayer mode.

2) For knockout mode, whenever "next challenger" beats "previous winner" we need to switch next Challenger to be "previous winner", but issue with output display is that it becomes "Player 4 v Player 4" if(player 1 has been winning but loses to player 4). 

-Variable Dice Mode, 
- Variable Player Mode. 

Comfort Level (1-5):
3 if 5 is most comfortable. 

Completeness Level (1-5):
3.5

Good learning experience and forces us to dig deep and think about thinking.

- Figuring out how to get variable Player mode to work automatically and show up all the combined numbers  I've always been weak at loops. 
- figuring out and implementing specific input validation to each gameStage (wrongly named as gameModes in the code).

- how to better organise and breakdown the different gameModes as will as the different Stages/Steps in the game. 
-  which one should be higher up in heirachy? 
Project 2 previous_current_Player_issue

### DIFF
--- a/script.js
+++ b/script.js
@@ -625,19 +625,19 @@ var inputQtyOfDice = function (numberOfDice) {
 var setMultiplayerOrKnockOut = function (playMode) {
   var displayPlayMode = "";
   var message2 = "";
-  var normalMessage = `Game will be played in ${displayPlayMode} style.
-                 <br>Please click "Submit" to proceed.`;
   var errorMessage = `XXX INPUT ERROR XXX
                       <br>Please input 'knock out' or 'multiplayer' only.`;
 
   if (playMode.toLowerCase() == "knock out") {
     gameMode = `arrangeDiceKnockOut`;
     displayPlayMode = "Knock Out";
-    message2 = normalMessage;
+    message2 = `Game will be played in ${displayPlayMode.toUpperCase()} style.
+                 <br>Please click "Submit" to proceed.`;
   } else if (playMode.toLowerCase() == "multiplayer") {
     gameMode = `arrangeDiceMultiplayer`;
     displayPlayMode = "Multiplayer";
-    message2 = normalMessage;
+    message2 = `Game will be played in ${displayPlayMode.toUpperCase()} style.
+                 <br>Please click "Submit" to proceed.`;
   } else {
     gameMode = `setMultiplayerOrKnockOut`;
     message2 = errorMessage;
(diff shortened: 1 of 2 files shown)


script.js
```diff
@@ -1,4 +1,943 @@
...
+  } else if (gameMode == `inputQtyOfDice`) {
+    myOutputValue = inputQtyOfDice(input);
+  } else if (gameMode == `setMultiplayerOrKnockOut`) {
+    myOutputValue = setMultiplayerOrKnockOut(input);
+  } else if (gameMode == `arrangeDiceMultiplayer`) {
+    myOutputValue = arrangeDiceMultiplayer(input);
+  } else if (gameMode == `arrangeDiceKnockOut`) {
+    myOutputValue = arrangeDiceKnockOut(input);
+  }
+
```
JustinWong98, line 941: You may want to look at using a switch case for a situation where you have multiple else ifs for the same condition:
https://www.w3schools.com/js/js_switch.asp

script.js
```diff
@@ -1,4 +1,943 @@
...
+  // use over arcing "IF /ELSE " to control for highest or lowest mode and return error messages
+  if (qtyOfPlayers > 1 && isNaN(qtyOfPlayers) == false) {
+    numberOfPlayers = qtyOfPlayers;
+    var message0 = `Number of Players = ${qtyOfPlayers}
+                  <br><br>Please input how many dices to use: `;
+    gameMode = `inputQtyOfDice`;
+  } else {
+    message0 = `xxxx INPUT ERROR xxxx :<br>Please key in only numerals (> 1)`;
+    gameMode = "getNumberOfPlayers";
+  }
```
JustinWong98, line 600: You won't need line 599 since you only change the gamemode when the first if statement runs, but overall this is good

script.js
```diff
@@ -1,4 +1,943 @@
...
+          diceRollsArraySorted = diceRollsArray.sort((a, b) => a - b);
+          diceRollsArraySortedString = diceRollsArraySorted.map(String);
+        }
+      }
+      //combining the diceRoll array to form number
+      combinedNumberString = diceRollsArraySortedString.join("");
+      combinedNumber = Number(combinedNumberString);
+
+      //Current Player Assignment + their scores
+      playerName = "Player " + (playerIndexNmbr + 1);
```
JustinWong98, line 723: Having this variable being named playerName isn't really accurate as it is more accurate to call it playerID

script.js
```diff
@@ -1,4 +1,943 @@
...
+
+        //sorting the diceRoll Array
+        if (gameLogic == "highest") {
+          diceRollsArraySorted = diceRollsArray.sort((a, b) => b - a);
+          diceRollsArraySortedString = diceRollsArraySorted.map(String);
+        } else if (gameLogic == "lowest") {
+          diceRollsArraySorted = diceRollsArray.sort((a, b) => a - b);
+          diceRollsArraySortedString = diceRollsArraySorted.map(String);
+        }
+      }
```
JustinWong98, line 817: Good use of the sort functionality!

script.js
```diff
@@ -1,4 +1,943 @@
...
+    gameMode = `arrangeDiceMultiplayer`;
+    displayPlayMode = "Multiplayer";
+    message2 = `Game will be played in ${displayPlayMode.toUpperCase()} style.
+                 <br>Please click "Submit" to proceed.`;
+  } else {
+    gameMode = `setMultiplayerOrKnockOut`;
+    message2 = errorMessage;
+  }
+
+  gameLogicDisplay = ` Game Logic: ${gameLogic.toUpperCase()} combined number wins<br><br>`;
```
JustinWong98, line 646: I think what you could do help cut down on some lines is to do the entire if else if else block and then define your message2 after that. Since message2 only changes with the name of the play mode, then you pl;ug the name of the variable afterwards. You can just return the errorMessage  on line 643 so that this situation works incase of an error

JustinWong98 (commented):
I didn't make a lot of comments, but I think overall your code is very complete - and generally well organised. I think the only thing to really focus on is seeing how you can refactor and take out redundant code - let me know if you wish for me to point out exactly where it is

PR #221, opened 2021-10-29
David Meng | Basics 11-7

davemeng:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

How many hours did you spend on this assignment? 2

Please fill in one error and/or error message you received while working on this assignment. Figuring out the right place to put the return. 

What part of the assignment did you spend the most time on? Proper syntax for concat and else/else if

Comfort Level (1-5): 4

Completeness Level (1-5): 4

What did you think of this deliverable? I probably need to refactor my code a lot more. 

Is there anything in this code that you feel pleased about? The (misnamed) rollTwoDice function but I haven't had time to use it properly for the More Comfortable exercises. 

What's one aspect of your code you would like specific, elaborate feedback on? Can I use concat without "" at the front, general code structure. What is the optimal position of the returns.

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,90 @@
+//Initiate variables and default game mode
+var gameMode = "P1RollDice";
+var P1Rolls = [];
+var P2Rolls = [];
+var P1Choice = 0;
+var P2Choice = 0;
+var numberOfDice = 2;
+
+//Function to roll x number of dice rolls
+var rollTwoDice = function (playerID) {
+  currentNumberOfDiceRolls = 0;
+  while (currentNumberOfDiceRolls < numberOfDice) {
+    var currentDiceRollNumber = Math.ceil(Math.random() * 6);
+    playerID[currentNumberOfDiceRolls] = currentDiceRollNumber;
+    currentNumberOfDiceRolls = currentNumberOfDiceRolls + 1;
+  }
+};
+
 var main = function (input) {
-  var myOutputValue = 'hello world';
+  var myOutputValue = "";
+  //P1 Roll Dice
+  if (gameMode == "P1RollDice") {
+    console.log("Current Game Mode:" + gameMode);
+    rollTwoDice(P1Rolls);
+    console.log("P1 Dice Rolls:" + P1Rolls);
+    myOutputValue = "You rolled a " + P1Rolls[0] + " for Dice 1 and a " + P1Rolls[1] + " for Dice 2.<br>Choose the order of the dice by selecting whether 1 (Dice 1) or 2 (Dice 2) should be the first digit.";
+    gameMode = "P1ChooseOrder";
+    console.log("Current Game Mode:" + gameMode);
+    return myOutputValue;
+  }
+  //P1 Choose Order of Dice
+  if (gameMode == "P1ChooseOrder") {
+    if (input == 1) {
+      console.log("P1 Choose Which Dice First:" + input);
+      myOutputValue = "Your number is ".concat(P1Rolls[0], P1Rolls[1]) + "<br> It is now Player 2's turn. ";
+      P1Choice = Number("".concat(P1Rolls[0], P1Rolls[1]));
+      gameMode = "P2RollDice";
+      console.log("P1Choice:" + P1Choice);
+    } else if (input == 2) {
+      console.log("P1 Choose Which Dice First:" + input);
+      myOutputValue = "Your number is ".concat(P1Rolls[1], P1Rolls[0]) + "<br> It is now Player 2's turn. ";
+      P1Choice = Number("".concat(P1Rolls[1], P1Rolls[0]));
+      gameMode = "P2RollDice";
+      console.log("P1Choice:" + P1Choice);
+    } else {
+      myOutputValue = input + " is an invalid selection. <br> Please choose the order of the dice by selecting whether 1 (Dice 1) or 2 (Dice 2) should be the first digit.";
+    }
+    return myOutputValue;
+  }
+  //P2 Roll Dice
+  if (gameMode == "P2RollDice") {
+    console.log("Current Game Mode:" + gameMode);
+    rollTwoDice(P2Rolls);
+    console.log("P2 Dice Rolls:" + P2Rolls);
+    myOutputValue = "You rolled a " + P2Rolls[0] + " for Dice 1 and a " + P2Rolls[1] + " for Dice 2.<br>Choose the order of the dice by selecting whether 1 (Dice 1) or 2 (Dice 2) should be the first digit.";
+    gameMode = "P2ChooseOrder";
+    console.log("Current Game Mode:" + gameMode);
+    return myOutputValue;
+  }
+  //P2 Choose Order of Dice
+  if (gameMode == "P2ChooseOrder") {
+    if (input == 1) {
+      console.log("P2 Choose Which Dice First:" + input);
+      myOutputValue = "Your number is ".concat(P2Rolls[0], P2Rolls[1]);
+      P2Choice = Number("".concat(P2Rolls[0], P2Rolls[1]));
+      console.log("P2Choice:" + P2Choice);
+      gameMode = "Choose Winner";
+      console.log("Current Game Mode:" + gameMode);
+    } else if (input == 2) {
+      console.log("P2 Choose Which Dice First:" + input);
+      myOutputValue = "Your number is ".concat(P2Rolls[1], P2Rolls[0]);
+      P2Choice = Number("".concat(P2Rolls[1], P2Rolls[0]));
+      console.log("P2Choice:" + P2Choice);
+      gameMode = "Choose Winner";
+      console.log("Current Game Mode:" + gameMode);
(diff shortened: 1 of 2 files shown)


script.js
```diff
@@ -1,4 +1,90 @@
...
+var gameMode = "P1RollDice";
+var P1Rolls = [];
+var P2Rolls = [];
+var P1Choice = 0;
+var P2Choice = 0;
+var numberOfDice = 2;
+
+//Function to roll x number of dice rolls
+var rollTwoDice = function (playerID) {
+  currentNumberOfDiceRolls = 0;
```
leechuanxin, line 11: You will have to initialise your variables with the `var` keyword. `var` scopes the creation of a variable to a function (if it's initialised within the function). Otherwise in this case without the `var`, JavaScript will assume you are initialising to the global scope instead.

Not initialising your variables with `var`, especially for the ones meant to be function scoped, may cause a chain of problems in the future where your variable gets accidentally modified. This is especially true the larger your program gets, and if you have identically named variables on the global scope and function scope.

You can learn more from this Stackoverflow comment here: https://stackoverflow.com/questions/2485423/is-using-var-to-declare-variables-optional/2485439#2485439

script.js
```diff
@@ -1,4 +1,90 @@
+//Initiate variables and default game mode
+var gameMode = "P1RollDice";
+var P1Rolls = [];
+var P2Rolls = [];
+var P1Choice = 0;
+var P2Choice = 0;
```
leechuanxin, line 6: Just a small reminder to use lower camel cases for naming your ordinary JavaScript variables: https://basics.rocketacademy.co/3-basic-data-manipulation/3.2-variables#variable-naming-convention. In this case, we will prefer something like `p1Rolls` instead of `P1Rolls`.

script.js
```diff
@@ -1,4 +1,90 @@
...
+    gameMode = "P1ChooseOrder";
+    console.log("Current Game Mode:" + gameMode);
+    return myOutputValue;
+  }
+  //P1 Choose Order of Dice with input validation
+  if (gameMode == "P1ChooseOrder") {
+    if (input == 1) {
+      console.log("P1 Choose Which Dice First:" + input);
+      myOutputValue = "Your number is ".concat(P1Rolls[0], P1Rolls[1]) + "<br> It is now Player 2's turn. ";
+      P1Choice = Number("".concat(P1Rolls[0], P1Rolls[1]));
```
leechuanxin, line 36: I am referring to the part which you needed specific feedback on, which is how you can combine the numbers as a string, and convert it back into a number again without using `"".concat(...)`.

There are a couple of ways to do that. Firstly, you can use template literals. Template literals, unlike your regular strings, do not begin and end with your regular quotation marks (single or double). Instead, they use the backtick symbol (the one to the left of your "!" key). Take for instance:

```
P1Choice = Number(`${P1Rolls[0]}${P1Rolls[1]}`)
```

The `${...}` syntax between the backticks allow you to pass JavaScript variables directly. In the example above, I am using `P1Rolls[0]` for the first, and `P1Rolls[1]` for the second.

Secondly, you can also combine array elements in a string using the `Array.join` method. You can look at the examples in the JavaScript demo here, just to have an understanding of the syntax: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join

For the case of line 36, I will probably do something like:
```
P1Choice = Number(P1Rolls.join(""))
```
Setting it up with the latter method (`Array.join`) may make it easier for you to embark on requirements such as [Variable Number of Dices](https://basics.rocketacademy.co/projects/project-2-beat-that#variable-number-of-dice), because you need not manually state the different dice indices like you are doing now. Instead, you can use `Array.join`, use the entire array, and not have to know in advance the number of elements/indices.

script.js
```diff
@@ -1,4 +1,90 @@
...
+  }
+  //P2 Roll Dice
+  if (gameMode == "P2RollDice") {
+    console.log("Current Game Mode:" + gameMode);
+    rollTwoDice(P2Rolls);
+    console.log("P2 Dice Rolls:" + P2Rolls);
+    myOutputValue = "You rolled a " + P2Rolls[0] + " for Dice 1 and a " + P2Rolls[1] + " for Dice 2.<br>Choose the order of the dice by selecting whether 1 (Dice 1) or 2 (Dice 2) should be the first digit.";
+    gameMode = "P2ChooseOrder";
+    console.log("Current Game Mode:" + gameMode);
+    return myOutputValue;
```
leechuanxin, line 58: Regarding your question about the optimal position of `return`s, I think the way you have set it up is alright. 

This is because you want to change the game mode like you see on [line 56](https://github.com/rocketacademy/basics-beat-that/blob/644677a86aec7d0e0b5470eccc2426a27879392c/script.js#L56), without going into the next `if` block (where we check if `gameMode` is `P2ChooseOrder`) right away.

With this setup, unfortunately we will need to `return` within the `if` blocks to terminate the function early and show the copywriting (`myOutputValue`) relevant to the current game state. 

Without the `return` it will proceed on to the next game state (`"P2ChooseOrder"`) immediately , and the player will never getting to see the copywriting for the `P2Rolls`: `"You rolled a " + P2Rolls[0] + " for Dice 1 and a " + P2Rolls[1] + " for Dice 2. ..."`

The alternative is using `else if` blocks for the different game states. Imagine:
```
if (gameMode == "P1RollDice") {
  ...
  // set the text of myOutputValue to whatever you want for this game state
  // no need to return since we are using if-else if chains
  myOutputValue = " ... ... ...";
} else if (gameMode == "P1ChooseOrder") {
  ...
  myOutputValue = " ... ... ...";
} else if (gameMode == "P2RollDice") {
  ...
  myOutputValue = " ... ... ...";
}
// ... and so on, and so forth for the rest of your game states ...
return myOutputValue;
```
This means every time the Submit button is clicked, it will only try to match one of the conditionals in the `(if)-(else if)` chain. With that, you do not need to do a `return myOutputValue` for each condition. You need only set it like so `myOutputValue = "...some string"`, and then just `return myOutputValue;` at the end of the `main` function.

Rather, if you are worried about the tidiness of your code, I will recommend you modularise each game state into a function. This is only assuming you want to continue using `if` blocks without using `else if` blocks for the game state conditions. Let's take this `P2RollDice` game state for instance. Won't your code be neater if you can do it this way:

```
if (gameMode == "P2RollDice") {
  return p2RollDice();
}

if (gameMode == "P2ChooseOrder) {
  return p2ChooseOrder();
}

// ... and so on, and so forth for the rest of the game states ...
```
where the logic in the `p2RollDice` function looks like this:

```
var p2RollDice = function () {
  // P2Rolls is a global variable, no need to be passed as param
  rollTwoDice(P2Rolls);
  // P2Rolls is a global variable, no need to be passed as param
  var output = "You rolled a " + P2Rolls[0] + " for Dice 1 and a " + P2Rolls[1] + " for Dice 2.<br>Choose the order of the dice by selecting whether 1 (Dice 1) or 2 (Dice 2) should be the first digit.";
  // gameMode is a global variable, no need to be passed as param
  gameMode = "P2ChooseOrder";
  return output;
};
```
If you want to modularise your game states into functions, together with using `else if`s, then the code will look like this instead:
```
if (gameMode == "P2RollDice") {
  myOutputValue = p2RollDice();
} else if (gameMode == "P2ChooseOrder") {
  myOutputValue = p2ChooseOrder();
}
// ... and so on and so forth ...
return myOutputValue;
```
And the function for p2RollDice should still look the same, like this:
```
var p2RollDice = function () {
  // P2Rolls is a global variable, no need to be passed as param
  rollTwoDice(P2Rolls);
  // P2Rolls is a global variable, no need to be passed as param
  var output = "You rolled a " + P2Rolls[0] + " for Dice 1 and a " + P2Rolls[1] + " for Dice 2.<br>Choose the order of the dice by selecting whether 1 (Dice 1) or 2 (Dice 2) should be the first digit.";
  // gameMode is a global variable, no need to be passed as param
  gameMode = "P2ChooseOrder";
  return output;
};
```

leechuanxin (commented):
Hey David,

Overall your game logic is fine. Just a couple of things to point out:

1. Be careful about you are initialising your variables. If you are creating a variable (as opposed to re-assigning values to an existing one), please use the `var` keyword. Things may feel fine with the smaller programs we are working with in Basics, but JavaScript may get _very confused_ for bigger programs, especially if you have multiple `.js` files embedded in 1 `.html` file, and we do not know if we have identically named variables in the other files.
2. Your game logic can be a lot neater by creating more functions, and wrapping your logic around these functions. For starters, you can split your main function up (while shortening it) to have it execute game state functions. You can review this video about [refactoring your game states with functions](https://basics.rocketacademy.co/8-managing-state-and-input-validation/8.2-program-state-for-game-modes#refactoring-code) on how you can achieve that.
3. EDIT: Missed that out earlier, but good job on trying out arrays and saving your dice numbers in one. It's definitely a lot cleaner storing dice rolls in arrays, compared to creating variables just to store each dice roll for each player.

PR #238, opened 2021-10-30
beat that - xiao you

amokxy:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

How many hours did you spend on this assignment? 2.5

logical errors working on while loops

figuring out the logic

Comfort Level (1-5): 3

Completeness Level (1-5): 3

What did you think of this deliverable? more challenging

Is there anything in this code that you feel pleased about? at least finished base? haha. wish i had more time for the more comfortable portions.

What's one aspect of your code you would like specific, elaborate feedback on? code logic and efficiency (eg can it be more concise)

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,124 @@
+// store game modes
+var gameMode = "start";
+var DICEROLLMODE = "dice roll mode";
+var ORDERMODE = "order mode";
+var ENTERNUMOFUSERS = "enter number of users";
+
+// arrays
+
+var allDiceRollsArray = [];
+var orderedNumberArray = [];
+var orderedNumberArrayforSorting = [];
+
+//counts
+var numOfPlayers = 0;
+var playerNumber = 1;
+var numDiceRollsOverall = 0;
+var diceRollsPerUser = 0;
+var numOfUsersWhoHavePlayed = 0;
+var diceRollsAllowedPerUser = 2;
+
+// dice roll function
+var rollDice = function () {
+  var randomIntegerUpToSix = Math.floor(Math.random() * 6) + 1;
+  return randomIntegerUpToSix;
+};
+
+//helper function to roll dice and store in array
+var rollAndStoreDice = function () {
+  var playerDiceRoll = rollDice();
+  allDiceRollsArray.push(playerDiceRoll);
+  return playerDiceRoll;
+};
+
+var outputWinner = function (sortingArray, copyOfSortingArray) {
+  sortingArray.sort();
+  var maxNumber = sortingArray.pop();
+  var matchToPlayerNum = copyOfSortingArray.indexOf(maxNumber);
+  var winningPlayer = matchToPlayerNum + 1;
+  return (
+    "The winner is player number " +
+    winningPlayer +
+    "! Click submit to restart game!"
+  );
+};
+
 var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
+  var myOutputValue = "";
+
+  // start mode. instructs user to enter total number of players.
+  if (gameMode == "start") {
+    gameMode = ENTERNUMOFUSERS;
+    return "Welcome. Please enter the number of players for this game";
+
+    // in this mode, user enters number of players. game mode switches to dice roll mode.
+  } else if (gameMode == ENTERNUMOFUSERS) {
+    numOfPlayers = input;
+    console.log("store no of players " + numOfPlayers);
+    gameMode = DICEROLLMODE;
+    return (
+      "There are " +
+      numOfPlayers +
+      " players in this game. Now, player 1 please click submit twice to roll two dice."
+    );
+  }
+  // entering dice roll mode. The loop ensures one game per user, with each game entails two dice rolls and the ordering of the dice rolls.
+
+  while (numOfUsersWhoHavePlayed < numOfPlayers) {
+    console.log(numOfUsersWhoHavePlayed, numOfPlayers);
+
+    // include one more round on top of max dice rolls allowed. in the first two round user rolls dice, in last round user orders the dice rolls
+    while (diceRollsPerUser < diceRollsAllowedPerUser + 1) {
+      //the following while loop rolls dice the num of times allowed per user, in this case 2.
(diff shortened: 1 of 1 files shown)


script.js
```diff
@@ -1,4 +1,124 @@
+// store game modes
+var gameMode = "start";
+var DICEROLLMODE = "dice roll mode";
+var ORDERMODE = "order mode";
+var ENTERNUMOFUSERS = "enter number of users";
```
lee-kimmixq, line 5: Good use of variables to store these strings! Do take note of naming conventions - these should be in snake case (i.e. `DICE_ROLL_MODE` etc)

script.js
```diff
@@ -1,4 +1,124 @@
...
+};
+
+//helper function to roll dice and store in array
+var rollAndStoreDice = function () {
+  var playerDiceRoll = rollDice();
+  allDiceRollsArray.push(playerDiceRoll);
+  return playerDiceRoll;
+};
+
+var outputWinner = function (sortingArray, copyOfSortingArray) {
```
lee-kimmixq, line 34: Good use of helper function! 👍 
The only time this function is called (at line 123), it is called `outputWinner(orderedNumberArray, orderedNumberArrayforSorting)`. Since the 2 arguments (`orderedNumberArray` and  `orderedNumberArrayforSorting`) are already global variables, you don't need to pass these variables through the argument in order to access them - you can just use the global variable names in your function! (ie you can replace `sortingArray` -> `orderedNumberArray` and `copyOfSortingArray` -> `orderedNumberArrayforSorting` in lines 35-44, and no need to put anything in the () after `function`.

script.js
```diff
@@ -1,4 +1,124 @@
...
+  sortingArray.sort();
+  var maxNumber = sortingArray.pop();
+  var matchToPlayerNum = copyOfSortingArray.indexOf(maxNumber);
```
lee-kimmixq, line 37: Good use of array methods to manipulate the arrays! 
Do take note that the `.sort()` method sorts by unicode value, which might differ from sorting by increasing numerical order (see [Compare Functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort))

script.js
```diff
@@ -1,4 +1,124 @@
...
+    gameMode = DICEROLLMODE;
+    return (
+      "There are " +
+      numOfPlayers +
+      " players in this game. Now, player 1 please click submit twice to roll two dice."
+    );
+  }
+  // entering dice roll mode. The loop ensures one game per user, with each game entails two dice rolls and the ordering of the dice rolls.
+
+  while (numOfUsersWhoHavePlayed < numOfPlayers) {
```
lee-kimmixq, line 67: While loops are usually used for a continuous loop (until a condition is met). Here we are breaking out of the while loop (with a `return` statement) every cycle that it runs, so the while loop is not really 'looping'. An `if` statement might be more appropriate in this situation! (same for lines 71 and 73)

script.js
```diff
@@ -1,4 +1,124 @@
...
+// arrays
+
+var allDiceRollsArray = [];
+var orderedNumberArray = [];
+var orderedNumberArrayforSorting = [];
+
+//counts
+var numOfPlayers = 0;
+var playerNumber = 1;
+var numDiceRollsOverall = 0;
```
lee-kimmixq, line 16: From what I understand, the purpose of this variable is to keep track of the number of dice rolls that has occurred - of which all the dice rolls are stored in the `allDiceRollsArray`. If you make use of the `.length` property of the `allDiceRollsArray`, this variable won't be necessary anymore. (less variables means less (re-)assigning of values, less probability of logic errors too!)

script.js
```diff
@@ -1,4 +1,124 @@
...
+          guess1 +
+          " and " +
+          guess2 +
+          ". Order the two guesses to form a two-digit number. The player with the highest two-digit number wins."
+        );
+      } else if (gameMode == ORDERMODE) {
+        console.log("entered order mode");
+
+        orderedNumberArray.push(input);
+        orderedNumberArrayforSorting.push(input);
```
lee-kimmixq, line 98: Good job pushing into 2 separate arrays so we can compare them!

script.js
```diff
@@ -1,4 +1,124 @@
...
+var allDiceRollsArray = [];
+var orderedNumberArray = [];
+var orderedNumberArrayforSorting = [];
+
+//counts
+var numOfPlayers = 0;
+var playerNumber = 1;
+var numDiceRollsOverall = 0;
+var diceRollsPerUser = 0;
+var numOfUsersWhoHavePlayed = 0;
```
lee-kimmixq, line 18: How does this variable relate to `numOfPlayers`? Could we use the same variable instead of having to track 2 variables?

script.js
```diff
@@ -1,4 +1,124 @@
...
+            orderedNumberArray[numOfUsersWhoHavePlayed - 1] +
+            ". It is now the next player's turn"
+          );
+      }
+    }
+    diceRollsPerUser = 0;
+    playerNumber = playerNumber + 1;
+    return "Hello player no. " + playerNumber + " click submit to play.";
+  }
+  return outputWinner(orderedNumberArrayforSorting, orderedNumberArray);
```
lee-kimmixq, line 123: To allow your game to start a new round without refreshing the browser, you might need to reinitialise some variables (`numOfUsersWhoHavePlayer`, `playerNumber`) to `0` to meet the while conditions

lee-kimmixq (commented):
Good job Xiao You!
I think you have fully understood the use of game modes, if-statements and really good job on using arrays in this project! Also good job on refactoring the `outputWinner()` function out of the main function 👍 

Some suggestions to improve on your code are:
1. Use `+=` - for example `playerNumber += 1` instead of `playerNumber = playerNumber +1`
2. Use [template literals](https://basics.rocketacademy.co/4-getting-started-with-code/4.2-our-first-program#output-formatting) to format your strings! It reduces the possibility of syntax error due to missing + signs or quotation marks (as well as makes your code shorter and more readable)
3. Relook on which part of the code runs in different situations - there is quite abit of nested if-statements in the main function - do we need all of them?

PR #242, opened 2021-10-30
Beat that Draft1

defragator:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

4 and counting 

mainly syntax errors 

Understanding arrays and ensuring my strings were free of syntax errors. spell checking(?)

Comfort Level (1-5):
1 - challenging
Completeness Level (1-5):
3.5
cool project 

completed 50% of it

cant think of anything at the moment

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,84 @@
+var playerNumber = 1;
+var currentGameMode = `waiting for Player 1 roll`;
+var myOutputValue = "";
+var P1randomDiceNumber1 = "";
+var P1randomDiceNumber2 = "";
+var P1results = "";
+var P2randomDiceNumber1 = "";
+var P2randomDiceNumber2 = "";
+var P2results = "";
+
 var main = function (input) {
-  var myOutputValue = 'hello world';
+  // Player 1
+  if (currentGameMode == `waiting for Player 1 roll`) {
+    P1randomDiceNumber1 = rollDice();
+    P1randomDiceNumber2 = rollDice();
+    myOutputValue = `Welcome Player 1. <br>
+    You rolled ${P1randomDiceNumber1} for Dice 1 and ${P1randomDiceNumber2} for Dice 2. <br>
+    Choose the order of the dice.`;
+    currentGameMode = `choose order of dice`;
+  } else if (currentGameMode == `choose order of dice`) {
+    if (input == `Dice 1`) {
+      P1results = P1randomDiceNumber1 * 10 + P1randomDiceNumber2;
+      currentGameMode = `waiting for Player 2 roll`;
+      myOutputValue = `Player 1, you chose Dice 1 first.<br>
+      Your number is ${P1results}.<br>
+      It is now Player 2's turn.`;
+    } else if (input == `Dice 2`) {
+      // store Dice 2 before Dice 1
+      P1results = P1randomDiceNumber2 * 10 + P1randomDiceNumber1;
+      currentGameMode = `waiting for Player 2 roll`;
+      myOutputValue = `Player 1, you chose Dice 2 first.<br>
+      Your number is ${P1results}.<br>
+      It is now Player 2's turn.`;
+    }
+  } else if (currentGameMode == `waiting for Player 2 roll`) {
+    P2randomDiceNumber1 = rollDice();
+    P2randomDiceNumber2 = rollDice();
+    myOutputValue = `Welcome Player 2. <br>
+    You rolled ${P2randomDiceNumber1} for Dice 1 and ${P2randomDiceNumber2} for Dice 2. <br>
+    Choose the order of the dice.`;
+    currentGameMode = `P2 choose order of dice`;
+  } else if (currentGameMode == `P2 choose order of dice`) {
+    if (input == `Dice 1`) {
+      P2results = P2randomDiceNumber1 * 10 + P2randomDiceNumber2;
+      // store Dice 1 before Dice 2
+      myOutputValue = `Player 2, you chose Dice 1 first.<br>
+      Your number is ${P2results}.<br>
+      Here are the results!`;
+      currentGameMode = `results`;
+    } else if (input == `Dice 2`) {
+      // store Dice 2 before Dice 1
+      P2results = P2randomDiceNumber2 * 10 + P2randomDiceNumber1;
+      myOutputValue = `Player 2, you chose Dice 2 first.<br>
+      Your number is ${P2results}.<br>
+      Here are the results!`;
+      currentGameMode = `results`;
+    }
+  } else if (currentGameMode == `results`) {
+    if (P1results > P2results) {
+      myOutputValue = `Player 1 wins!<br>
+      Player 1 rolled ${P1results}.<br>
+      Player 2 rolled ${P2results}`;
+    }
+    if (P2results > P1results) {
+      myOutputValue = `Player 2 wins! <br>
+      Player 1 rolled ${P1results}.<br>
+      Player 2 rolled ${P2results}`;
+    }
+  }
   return myOutputValue;
+};
+
+var rollDice = function () {
+  // produces a decimal between 0 and 6
+  var randomDecimal = Math.random() * 6;
+
+  // take off the decimal
+  var randomInteger = Math.floor(randomDecimal);
+
+  // it's anumber from 0 - 5 ... add 1
+  var diceNumber = randomInteger + 1;
+
+  return diceNumber;
 };


defragator:
Draft 1

script.js
```diff
@@ -1,4 +1,84 @@
+var playerNumber = 1;
+var currentGameMode = `waiting for Player 1 roll`;
```
leechuanxin, line 2: Good attempt at trying out template literals (using ``waiting for Player 1 roll`` instead of `'waiting for Player 1 roll'`) by Project 2! 

You will find that as you learn JavaScript further, template literals will make string modifications a lot easier. This is because the syntax will be less verbose: compare ``waiting for Player ${playerNum} roll`` instead of `'waiting for Player ' + playerNum + ' roll'`.

I will recommend saving the game mode strings as global constants, so that you can only refer to the variables throughout your file. What I meant is:

```
// for global constants, we use scream-casing for naming variables
// full capitalisation, with underscores to separate words
// "constants" meaning we don't expect to ever change the words in the strings manually,
// nor through any code manipulating the string
var WAITING_FOR_PLAYER1_ROLL = `waiting for Player 1 roll`;
var CHOOSE_ORDER_OF_DICE = `choose order of dice`;
// set default game mode on page load to be WAITING_FOR_PLAYER1_ROLL
var currentGameMode = WAITING_FOR_PLAYER1_ROLL;
...
...
if (currentGameMode == WAITING_FOR_PLAYER1_ROLL) {
  ...
  currentGameMode = CHOOSE_ORDER_OF_DICE;
}
...
else if (currentGameMode == CHOOSE_ORDER_OF_DICE) {
  ...
}
```
Notice this is different from your current method of checking `currentGameMode` directly against a string, like so `if (currentGameMode == `waiting for Player 1 roll`) { ... }`. 

Saving these strings as variables early in your script keeps your code DRY (acronym for "Don't Repeat Yourself"), and allows you to make changes only when first assigning the variables. If we were to capitalise the first letter of `"Waiting"` in your string `"waiting for Player 1 roll"`, notice that 
we would have to change it throughout the script in quite a few places if we were to refer to the string (instead of the variables) directly?

script.js
```diff
@@ -1,4 +1,84 @@
...
+var myOutputValue = "";
+var P1randomDiceNumber1 = "";
+var P1randomDiceNumber2 = "";
+var P1results = "";
+var P2randomDiceNumber1 = "";
+var P2randomDiceNumber2 = "";
+var P2results = "";
```
leechuanxin, line 9: Arrays and looping with arrays have already been covered at the time of submission, and it will make your storage of data simpler.

Here's a look at the data structures you could have employed, without the need to declare so many global variables.

```
// array indices (indexes) count from 0,
// so we can take `0` to refer to Player 1, and
// `1` to refer to Player 2.
var currentPlayerTurn = 0;
// for the sake of simplicity, let's assume I am only
// having 1 array just for storing results. They are
// empty strings now. To get P1results, it will be
// the equivalent of results[0], and for P2results,
// it will be results[1].
var results = ["", ""];
```
With a `currentPlayerTurn` initialised, `waiting for Player 1 roll` and `waiting for Player 2 roll` need not be 2 separate game modes. Likewise for `choose order of dice`, you can keep to 1 game mode. You can simply have 1 `waiting for player roll` game mode, and then switch to the other player once `choose order of dice` is done for any player. This also means instead of separate game modes for player 1 and player 2, now you need to be able to switch between players.

Consider the following pseudocode logic:
```
...
// currentGameMode is `choose order of dice`
  // we check if it's Player 1's turn
  // if it is, and when the rest of the logic is executed, move to player 2
  // you can do so using `currentPlayerTurn += 1;`
  // else, it is Player 2's turn
    // given your game does not have a reset functionality, when it's player 2's turn,
    // you switch to `results` game mode instead of switching to player 1
    // if it's Player 2 (last player's turn) and you intend to have a game reset functionality,
    // change player to player1, like so `currentPlayerTurn = 0;`
```

script.js
```diff
@@ -1,4 +1,84 @@
+var playerNumber = 1;
+var currentGameMode = `waiting for Player 1 roll`;
+var myOutputValue = "";
```
leechuanxin, line 3: I wouldn't recommend setting `myOutputValue` as a global variable. It need not be modified nor checked against any other function, and it's used only to be `return`-ed by `main`. It can be kept and initialised in `main`.

script.js
```diff
@@ -1,4 +1,84 @@
...
 var main = function (input) {
-  var myOutputValue = 'hello world';
+  // Player 1
+  if (currentGameMode == `waiting for Player 1 roll`) {
+    P1randomDiceNumber1 = rollDice();
+    P1randomDiceNumber2 = rollDice();
+    myOutputValue = `Welcome Player 1. <br>
+    You rolled ${P1randomDiceNumber1} for Dice 1 and ${P1randomDiceNumber2} for Dice 2. <br>
+    Choose the order of the dice.`;
+    currentGameMode = `choose order of dice`;
```
leechuanxin, line 19: By Project 2, you can begin to make your `main` function more succinct by moving game mode conditional logic into their own functions.

Assuming we are using arrays as previously recommended to store `results`, we can do so this way:
```
// we assume playerTurn counts from 0
// so 0 for player 1, 1 for player 2
var rollForPlayer = function (playerTurn) {
  // we add an empty string in front so that JavaScript
  // knows we are combining strings instead of adding numbers
  var diceRollString = "" + rollDice() + rollDice();
  results[playerTurn] = diceRollString;
  // assume we roll "1" and "3", diceRollString will be "13"
  // to get "1" and "3" separately again from diceRollString,
  // we can do so this way: results[playerTurn][0] and results[playerTurn][1]
  // this means, get the first character of results[playerTurn] and its second character:
  // giving us "1" and "3" separately.
  var output = `Welcome Player ${playerTurn + 1}. <br>
    You rolled ${results[playerTurn][0]} for Dice 1 and ${results[playerTurn][1]} for Dice 2. <br>
    Choose the order of the dice.`;
  currentGameMode = CHOOSE_ORDER_OF_DICE;
  return output;
}

var main = function (input) {
  var myOutputValue = "";
  if (currentGameMode == WAITING_FOR_PLAYER_ROLL) {
    myOutputValue = rollForPlayer(currentPlayerTurn);
  }
  ...
}
```

script.js
```diff
@@ -1,4 +1,84 @@
...
+      Your number is ${P1results}.<br>
+      It is now Player 2's turn.`;
+    } else if (input == `Dice 2`) {
+      // store Dice 2 before Dice 1
+      P1results = P1randomDiceNumber2 * 10 + P1randomDiceNumber1;
+      currentGameMode = `waiting for Player 2 roll`;
+      myOutputValue = `Player 1, you chose Dice 2 first.<br>
+      Your number is ${P1results}.<br>
+      It is now Player 2's turn.`;
+    }
```
leechuanxin, line 34: Assuming again we are following the same logic recommended of just storing the result in an array, you do not need to do the multiplication method here. In fact, only 1 conditional is required.

```
// dice 1 not required, because the previous logic
// of storing results right away already assumes dice 1 first, then dice 2.
if (input == `Dice 2`) {
  // in this case, we are splitting the 2 dice rolls, 
  // putting the 2nd one in front and the 1st at the back
  // then recombining them using string concatenation to be
  // stored back in results
  results[currentPlayerTurn] = results[currentPlayerTurn][1] + results[currentPlayerTurn][0];
  currentGameMode = ...
  myOutputValue = ...
}
```

leechuanxin (commented):
Thanks for the submission Shane. Overall, good attempt. There's quite a bit of repetition in the logic though, namely:

- If you are using arrays and counters to store results and current player turn, there's no need to have multiples of the same game modes for different players.
- I understand this is strictly for more comfortable attempts, but you can also simply just have an array variable to store results, instead of having separate ones to store dice rolls, and then results. By default, we can assume the results are going to be storing dice roll 1 first, then dice roll 2. Then we only switch the order using 2D array syntax (someArray[index1][index2]) and `+` as a string concatenation operator when the user requests for dice 2 to be the first dice number, like so: `  results[currentPlayerTurn] = results[currentPlayerTurn][1] + results[currentPlayerTurn][0];`
- `main` function can be kept shorter by beginning to move some of the game mode conditional logic into their own functions.

For future pull requests attempts for Rocket Academy submissions, it will be good to title your pull requests "Basics [batchNumber] Section [sectionNumber] [yourName]". This pull request is currently titled "Beat that Draft1", and we haven't been able to find your pull requests; we couldn't match the username `defragator` to Shane :P I guess that's the reason we have missed of your pull requests long ago :)

PR #226, opened 2021-10-29
Yuting's Project 2

CheongYuting:
Please fill out the survey before submitting the pull request. Thanks!

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

How many hours did you spend on this assignment?
>10 hours

Please fill in one error and/or error message you received while working on this assignment.
Unable to restart the round for player 2. Learned that its because i forgot to reset the array for the dice numbers rolled. 
This mistake was repeated for restarting the game (forgot to reset the array for players' scores).

What part of the assignment did you spend the most time on?
Debugging

Comfort Level (1-5):
1

Completeness Level (1-5):
4

What did you think of this deliverable?
Needs more formatting to make it complete but it was a good challenge, makes me think

Is there anything in this code that you feel pleased about?

What's one aspect of your code you would like specific, elaborate feedback on?
How do I format the code so that when its player 2, I can change the output message to say "Click submit to see who wins"?

PR #255, opened 2021-11-02
11-7

yuqinggg:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

4-5h

_"undefined playerNumber"_

Assigning Player 1 and Player 2

Comfort Level (1-5):
4

Completeness Level (1-5):
4.5

Mind-boggling

It only uses one button but have so many functions :D

Ordering of my codes (e.g. main functions, variables, global functions, etc) I think such good practices are important as we are starting to code more complicated things.

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,117 @@
+// Define current player
+var currentPlayer = 1;
+
+// Function for output
+var myOutputValue = "";
+
+// Define variables to keep track of player's number
+var playerNumber = [];
+var playerOneNum = playerNumber[0];
+var playerTwoNum = playerNumber[1];
+
+// Function to track game mode
+var modeRollDice = "Roll dice mode";
+var modeChoseOrder = "Chose order of numbers";
+var modeCurrent = modeRollDice;
+
+// Function to keep track of dice rolled
+var diceNumberRolled = [];
+var diceRollA = "";
+var diceRollB = "";
+
+// Combination of numbers
+// var combineNumbers = "";
+var combineNumbersAB = Number(diceRollA) + Number(diceRollB);
+var combineNumbersBA = Number(diceRollB) + Number(diceRollA);
+
+// Function to get random dice number
+var rollDice = function () {
+  var randomDecimal = Math.random() * 6;
+  var randomInteger = Math.floor(randomDecimal);
+  var diceNumber = randomInteger + 1;
+  return diceNumber;
+};
+
 var main = function (input) {
-  var myOutputValue = 'hello world';
+  // Select game mode
+  var gameCounter = 0;
+  if (modeCurrent == modeRollDice) {
+    myOutputValue = rollDiceTwice(gameCounter);
+    modeCurrent = modeChoseOrder;
+  }
+  // Choose order of numbers
+  else if (modeCurrent == modeChoseOrder) {
+    // If Player chooses AB arrangement
+    var combinedNumbers = String(diceRollA) + String(diceRollB);
+    var myOutputValue = `The number you've chose is ${combinedNumbers} <br><br> Click 'submit' to continue. `;
+
+    // If Player chooses BA arrangement
+    if (input == diceRollB) {
+      var combinedNumbers = String(diceRollB) + String(diceRollA);
+      var myOutputValue = `The number you've chose is ${combinedNumbers} <br><br> Click 'submit' to continue. `;
+    }
+    // Function to combine players' numbers
+    playerNumber.push(combinedNumbers);
+    console.log(`Player ${Number(currentPlayer)} chose ${combinedNumbers}`);
+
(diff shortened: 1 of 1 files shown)


script.js
```diff
@@ -1,4 +1,117 @@
+// Define current player
+var currentPlayer = 1;
+
+// Function for output
+var myOutputValue = "";
```
leechuanxin, line 5: I strongly suggest keeping `myOutputValue` inside the `main` function, because it is the only function that `return`s the string back to the grey output box for rendering. You can view my comments to [Yuting's pull request](https://github.com/rocketacademy/basics-beat-that/pull/226#pullrequestreview-793589735) further on why, and how to achieve it.

script.js
```diff
@@ -1,4 +1,117 @@
...
+
+// Define variables to keep track of player's number
+var playerNumber = [];
+var playerOneNum = playerNumber[0];
+var playerTwoNum = playerNumber[1];
```
leechuanxin, line 10: These 2 unused variables can be removed. That said, if you do intend to use them, I suggest not assigning them to `playerNumber[0]` or `playerNumber[1]` on creation / initialisation. 

This is because `playerNumber` has no items as of its creation on line 8. It is initialised to be an empty array in [line 8](https://github.com/rocketacademy/basics-beat-that/blob/e3e941a6585a42407eb2a4218d945119427bf058/script.js#L8), so it means `console.log(playerOneNum)` or `console.log(playerTwoNum)` are going to give you `undefined` for both.

If you want them initialised to store a number, why not start with `0`? `0` does not make sense in the context of the application (dice rolls start from `1` and end with `6`), so at least you have some placeholder number.

script.js
```diff
@@ -1,4 +1,117 @@
...
+  var diceNumber = randomInteger + 1;
+  return diceNumber;
+};
+
 var main = function (input) {
-  var myOutputValue = 'hello world';
+  // Select game mode
+  var gameCounter = 0;
+  if (modeCurrent == modeRollDice) {
+    myOutputValue = rollDiceTwice(gameCounter);
```
leechuanxin, line 39: `myOutputValue` here will be using the version in the global scope, while the version in [line 46](https://github.com/rocketacademy/basics-beat-that/blob/e3e941a6585a42407eb2a4218d945119427bf058/script.js#L46) is in the function scope.

How about creating `myOutputValue` only within the `main` function so that it is scoped here? You can create it within the function outside of the `if` blocks that check for the game states, such as [below line 35](https://github.com/rocketacademy/basics-beat-that/blob/e3e941a6585a42407eb2a4218d945119427bf058/script.js#L36).

You can go `var myOutputValue = ...` below line 35, and just re-assigning strings to it to change its strings.

script.js
```diff
@@ -1,4 +1,117 @@
...
+    modeCurrent = modeChoseOrder;
+  }
+  // Choose order of numbers
+  else if (modeCurrent == modeChoseOrder) {
+    // If Player chooses AB arrangement
+    var combinedNumbers = String(diceRollA) + String(diceRollB);
+    var myOutputValue = `The number you've chose is ${combinedNumbers} <br><br> Click 'submit' to continue. `;
```
leechuanxin, line 46: I don't suggest creating variables within conditional blocks like `if`s, `else`s or `else if`s, unless you only intend to use them in said blocks. View my comment above for what I suggest instead.

script.js
```diff
@@ -1,4 +1,117 @@
...
 var main = function (input) {
-  var myOutputValue = 'hello world';
+  // Select game mode
+  var gameCounter = 0;
+  if (modeCurrent == modeRollDice) {
+    myOutputValue = rollDiceTwice(gameCounter);
+    modeCurrent = modeChoseOrder;
+  }
+  // Choose order of numbers
+  else if (modeCurrent == modeChoseOrder) {
```
leechuanxin, line 43: The logic inside this game state can be modularised into a function, similar to what you are already doing with [execution of the `rollDiceTwice` function, where you change the myOutputValue to the string returned from the function](https://github.com/rocketacademy/basics-beat-that/blob/e3e941a6585a42407eb2a4218d945119427bf058/script.js#L102-L117). As you can see [here](https://github.com/rocketacademy/basics-beat-that/blob/e3e941a6585a42407eb2a4218d945119427bf058/script.js#L43-L72), the end result of the function (or its `return`) can simply be a string that is `return`ed back to the grey output box.

leechuanxin (commented):
Hey Yuqing,

Thanks for submitting your pull request. Your style of writing (and subsequently, the mistakes) is similar to Yuting's, so [you can also view my comments to her for reference](https://github.com/rocketacademy/basics-beat-that/pull/226#pullrequestreview-793589735).

PR #275, opened 2021-12-11
12-4 Annabel Beat That

wolffsteinn:
…lay multiple times but did not work

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

4 hours

I tried to make the game to be playable multiple times between two players, the game allowed for 2 rounds but after that it will not reach the last block of code to compare the values to see which player won 

troubleshooting the loop of the game

Comfort Level (1-5):
4.5

Completeness Level (1-5):
4.5??

not satisfied with it. didnt have enough time to properly go through my codes also

managed to do a loop of the game at least once

was there any other helper functions that i can do to make the main function shorter? 
the part on the infinite loop... is there something that i am missing?

### DIFF
--- a/index.html
+++ b/index.html
@@ -108,8 +108,8 @@ The <html> element is known as the root element of a HTML page. All other elemen
       </p>
       <p>
         <center>
-          For example, if you roll a 2 and 5, the highest number possible would
-          be 52
+          For example, if you roll a 1 and 5, the highest number possible would
+          be 51
         </center>
       </p>
       <p>
--- a/index.html
+++ b/index.html
@@ -1,7 +1,15 @@
+<!--  All HTML files begin with the following line: -->
 <!DOCTYPE html>
+<!-- This indicates that the rest of the lines that follow are to be interpreted as HTML. It tells the browser about what document type to expect. -->
+
+<!-- All text enclosed in angled brackets are known as HTML element tags. There are usually, but not always, starting and closing element "tags", each representing a HTML element.
+The <html> element is known as the root element of a HTML page. All other elements are encased within it. -->
 <html>
+  <!-- The <head> element contains meta information about the HTML page, information that is not rendered or displayed by the browser -->
   <head>
-    <title>Coding Basics</title>
+    <!-- The <title> element specifies a title for the HTML page (which is shown in the browser's title bar or in the page's tab) and is the default name given to a bookmark of a webpage. -->
+    <title>Basics</title>
+    <!-- The <style> element contains information relating to the layout and style of the webpage. The syntax within a <style> tag is referred to as CSS - Cascading Style Sheets. -->
     <style>
       * {
         box-sizing: border-box;
@@ -18,10 +26,10 @@
       }
 
       #container {
-        background-color: pink;
-        margin: 40px auto;
-        max-width: 800px;
-        padding: 38px 31px;
+        background-color: rgb(125, 202, 216);
+        margin: 120px auto;
+        max-width: 1000px;
+        padding: 40px 40px;
       }
 
       input {
@@ -40,7 +48,7 @@
       }
 
       #output-div {
-        background-color: lightgrey;
+        background-color: rgba(241, 232, 190, 0.521);
         margin: 20px auto;
         padding: 20px;
         width: 100%;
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -40,32 +48,91 @@
...
+        <center>
+          For example, if you roll a 1 and 5, the highest number possible would
+          be 51
+        </center>
+      </p>
+      <p>
+        <center>
+          The player with the highest arranged number wins, so grab a buddy and
+          let's play! 😝
+        </center>
```
shannonssd, line 119: Great job playing around with the HTML file and adding instructions.
However, from my knowledge the center tags: <'center'>, is deprecated.
A better way to centralise text is to style it using CSS: https://www.w3schools.com/html/tryit.asp?filename=tryhtml_styles_text-align

What deprecated code is: Deprecated means that it is still in use, but only for historical purposes and it will be removed probably in the next big release. It is recommended that you do not use deprecated functions or features - even if they are present in the current library for example.

moreComfortable.js
```diff
@@ -0,0 +1,114 @@
+//two players will take turns to play
+//when the player clicks submit, game rolls 2 dice and show the values
+//player then can pick the order of dice they want ie: 36 or 63
+//after both players have rolled and chosen the dice order, the player with the highest combined number wins.
+//have to store the values of the chosen combination
```
shannonssd, line 5: Good job using pseudocode to plan what need to happen before writing code. Keep up this habit!

moreComfortable.js
```diff
@@ -0,0 +1,114 @@
...
+//since there will be two players, multiple program states needed
+var currentGameMode = "waiting for player 1 input";
+//rolling dice function
+var numberGenerator = function () {
+  var randomDecimal = Math.random() * 6;
+  var randomInteger = Math.floor(randomDecimal);
+  var randomNumber = randomInteger + 1;
+  return randomNumber;
+};
+//we can use loops to store the dice roll values into the array
```
shannonssd, line 20: Great job using comments to describe what each function is doing!

moreComfortable.js
```diff
@@ -0,0 +1,114 @@
...
+var dicePlayer1 = function () {
```
shannonssd, line 21: The convention is for functions to start with a verb where possible. So instead of dicePlayer1() a better name would be generatePlayer1Dice(). This also makes it easier for us to differentiate functions from variables which reduces the chances of human error when writing code.

moreComfortable.js
```diff
@@ -0,0 +1,114 @@
...
+    currentGameMode = "waiting for player 2 to choose dice order";
+
+    return myOutputValue;
+  }
+  if (currentGameMode == "waiting for player 1 to choose dice order") {
+    //have to concatenate the two values together
+    if (input == "1") {
+      currentGameMode = "waiting for player 2 input";
+      var player1 = Number(String(diceValues[0]) + String(diceValues[1]));
+      player1Choice[0] = player1Choice[0] + player1;
```
shannonssd, line 59: You do not have to assign var player1Choice = [0] at line 1 and then add player1's choice.
A better way would be to declare player1Choice as an empty array at line 1 and then use the .push method to add player 1's score to the array over here. 

.push(): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push

moreComfortable.js
```diff
@@ -0,0 +1,114 @@
...
+
+      return `This is your combined number: ${player1} and its time for Player 2's turn!`;
+    } else if (input == "2") {
+      currentGameMode = "waiting for player 2 input";
+      var player1 = Number(String(diceValues[1]) + String(diceValues[0]));
+      player1Choice[0] = player1Choice[0] + player1;
+
+      return `This is your combined number: ${player1} and its time for Player 2's turn!`;
+    } else if (input != "1" && input != "2") {
+      return "Please only enter 1 or 2!!";
```
shannonssd, line 69: Good use of input validation!

moreComfortable.js
```diff
@@ -0,0 +1,114 @@
...
+    }
+  }
+  if (currentGameMode == "waiting for player 2 to choose dice order") {
+    //have to concatenate the two values together
+    if (input == "1") {
+      currentGameMode = "comparing dice numbers";
+      var player2 = Number(String(diceValues[2]) + String(diceValues[3]));
+      player2Choice[0] = player2Choice[0] + player2;
```
shannonssd, line 77: Same as above. We should be using .push() here to add a value to an array

moreComfortable.js
```diff
@@ -0,0 +1,114 @@
...
+      );
+    } else if (Number(player1Choice) < Number(player2Choice)) {
+      diceValues = [];
+
+      currentGameMode = "waiting for player 1 input";
+      return (
+        "Player 2 won! <br> Click 'submit' to play the game again! <br>Player 2 score: " +
+        player2Choice[0] +
+        "<br>Player 1 score: " +
+        player1Choice[0]
```
shannonssd, line 110: This could be in a helper function outside of the main function.

shannonssd (approved):
Keep up the consistency of writing pseudocode and comments throughout your code.
Also good job using helper functions throughout the code. 

Keep up the great work Annabel! 😊

PR #284, opened 2021-12-11
12-2 Raymond

rayjosong:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

2.5h 

Getting undefined functions quite often. Often because I failed to return the code. 

Input validation

Comfort Level (1-5):
4

Completeness Level (1-5):
4

Great, because it forces me to combine and understand all concepts taught so far

User-friendliness

How it could be refactored further

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,7 +1,15 @@
+<!--  All HTML files begin with the following line: -->
 <!DOCTYPE html>
+<!-- This indicates that the rest of the lines that follow are to be interpreted as HTML. It tells the browser about what document type to expect. -->
+
+<!-- All text enclosed in angled brackets are known as HTML element tags. There are usually, but not always, starting and closing element "tags", each representing a HTML element.
+The <html> element is known as the root element of a HTML page. All other elements are encased within it. -->
 <html>
+  <!-- The <head> element contains meta information about the HTML page, information that is not rendered or displayed by the browser -->
   <head>
-    <title>Coding Basics</title>
+    <!-- The <title> element specifies a title for the HTML page (which is shown in the browser's title bar or in the page's tab) and is the default name given to a bookmark of a webpage. -->
+    <title>Basics</title>
+    <!-- The <style> element contains information relating to the layout and style of the webpage. The syntax within a <style> tag is referred to as CSS - Cascading Style Sheets. -->
     <style>
       * {
         box-sizing: border-box;
@@ -18,7 +26,7 @@
       }
 
       #container {
-        background-color: pink;
+        background-color: rgb(29, 240, 229);
         margin: 40px auto;
         max-width: 800px;
         padding: 38px 31px;
@@ -48,24 +56,44 @@
     </style>
   </head>
 
+  <!-- The <body> element defines the document's body, and is a container for all the visible contents that will be rendered by the browser.  -->
   <body>
-    <h1 id="header">Basics: Beat That! 🚀</h1>
+    <!-- The <h1> element defines a large heading. There are 6 pre-set header elements, <h1> to <h6> -->
+    <h1 id="header"><b>BEAT THAT😡🥊🤕🥲</b></h1>
+    <!-- The <div> tag defines a division or a section in an HTML document, and is commonly use as a "container" for nested HTML elements. -->
     <div id="container">
+      <p>
+        <b>
+          Hello! Welcome to BEAT THAT! Click submit to start the game.
+          Disclaimer: I know you just had a stressful day, but we are not
+          actually beating anyone.</b
+        >
+      </p>
+      <!-- The <p> tag defines a paragraph element. -->
       <p>Input:</p>
+      <!-- The self-closing <input /> tag represents a field for a user to enter input. -->
       <input id="input-field" />
+      <!-- The self-closing <br /> tag represents a line break - a line's worth of spacing before dispalying the next element. -->
       <br />
+      <!-- The <button> tag represents.. you guessed it! a button. -->
       <button id="submit-button">Submit</button>
       <p>Output:</p>
       <div id="output-div"></div>
     </div>
-    <!-- Import program logic -->
+    <!-- Everything below this is incorporating JavaScript programming logic into the webpage. -->
+    <!-- The script tag encloses JavaScript syntax for the browser to interpret programtically -->
+    <!-- The following line imports all code written in the script.js file -->
     <script src="script.js"></script>
     <!-- Define button click functionality -->
     <script>
+      // Declare and define a variable that represents the Submit Button
       var button = document.querySelector("#submit-button");
+      // When the submit button is clicked, access the text entered by the user in the input field
+      // And pass it as an input parameter to the main function as defined in script.js
       button.addEventListener("click", function () {
         // Set result to input value
         var input = document.querySelector("#input-field");
+        // Store the output of main() in a new variable
         var result = main(input.value);
 
         // Display result in output element
--- a/script.js
+++ b/script.js
@@ -1,4 +1,210 @@
+/*
+Requirements (Base)
+1. There are 2 players and players take turns 
+
+2. When a player clicks Submit, the game rolls 2 dice and shows the dice rolls, for example 3 and 6.
+3. The player picks the order of the dice they want. For example, if they wanted the number 63, they would specify that the 2nd dice goes first. You can choose how the player specifies dice order.
+4. After both players have rolled and chosen dice order, the player with the higher combined number wins.
+
+
+concepts involved
+1. Game states
+2. Loops
+3. Arrays 
+4. Loops w Arrays?
+5. Can Boolean be used? 
+
+
+how to 
+- must have mechanism to move from player 1 to player 2
+- 
+*/
+
+// Create variables for the game modes
+var player1Roll = "player 1 roll dices";
+var player1Select = "player 1 select dice order";
+
+var player2Roll = "player 2 roll dices";
+var player2Select = "player 2 select dice order";
+
+var matchPlayer1Player2 = "Check who wins";
+
+var player1FirstRollOutput = 0;
+var player1SecondRollOutput = 0;
+var player2FirstRollOutput = 0;
+var player2SecondRollOutput = 0;
+
+var player1Array = [];
+var player2Array = [];
+
+var player1Number = 0;
+var player2Number = 0;
+
+// set to true if the game is over
+var isTheGameOver = false;
+
+var selectDiceOrderText =
+  "Choose the order of the dice. Type '1' for first dice first. Alternatively, type '2' for second dice first.";
+
+// Set current game mode so user will roll dice when click submit
+var currentGameMode = player1Roll;
+
+//create function for dice roll
+var diceRoll = function () {
+  return Math.floor(Math.random() * 6) + 1;
+};
+
+//create function emcompassing all game modes & game logic
+var beatThatGame = function (input) {
+  // game mode: player 1 to roll
+  if (currentGameMode == player1Roll) {
+    if (input != "") {
+      myOutputValue = "Please roll the dice by clicking 'submit'";
+      return myOutputValue;
+    }
+    player1FirstRollOutput = diceRoll();
+    player1SecondRollOutput = diceRoll();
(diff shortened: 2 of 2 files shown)


index.html
```diff
@@ -48,24 +56,44 @@
...
+    <h1 id="header"><b>BEAT THAT😡🥊🤕🥲</b></h1>
+    <!-- The <div> tag defines a division or a section in an HTML document, and is commonly use as a "container" for nested HTML elements. -->
     <div id="container">
+      <p>
+        <b>
+          Hello! Welcome to BEAT THAT! Click submit to start the game.
+          Disclaimer: I know you just had a stressful day, but we are not
+          actually beating anyone.</b
+        >
+      </p>
```
lim-jiahao, line 71: Nice attempt at changing the html and css! I died at this message LOL. I like that you give instructions to the user on how to start the game too. A very common comment I have for other projects is that when I enter the page, I just see a blank input field and output box and it is unclear how to proceed. Good that you're taking the user experience into account

script.js
```diff
@@ -1,4 +1,210 @@
...
+var player2Number = 0;
+
+// set to true if the game is over
+var isTheGameOver = false;
+
+var selectDiceOrderText =
+  "Choose the order of the dice. Type '1' for first dice first. Alternatively, type '2' for second dice first.";
+
+// Set current game mode so user will roll dice when click submit
+var currentGameMode = player1Roll;
```
lim-jiahao, line 50: Nice use of global variables to track your game state and player rolls! For those that will stay constant such as the game states and the selectDiceOrderText, consider using the `const` keyword to declare them rather than var. For constants, some people will use SCREAM_CASE as well for the variable name. For example `const PLAYER_1_ROLL = "player 1 roll dices"`. I know this wasn't really covered in the course but it's a minor step up if you're interested.

For more info:

script.js
```diff
@@ -1,4 +1,210 @@
...
+  return Math.floor(Math.random() * 6) + 1;
+};
+
+//create function emcompassing all game modes & game logic
+var beatThatGame = function (input) {
+  // game mode: player 1 to roll
+  if (currentGameMode == player1Roll) {
+    if (input != "") {
+      myOutputValue = "Please roll the dice by clicking 'submit'";
+      return myOutputValue;
```
lim-jiahao, line 63: good input validation, i know we taught the use of the variable `myOutputValue` at the start but this was just to make it easy to understand what the variable is for.  in simple cases like this, you can actually just return the string directly so this can be simplified to `return "Please roll the dice by clicking 'submit'"`

script.js
```diff
@@ -1,4 +1,210 @@
...
+    }
+    player1FirstRollOutput = diceRoll();
+    player1SecondRollOutput = diceRoll();
+    player1Array.push(player1FirstRollOutput);
+    console.log("pushed player1firstrolloutput into player1array");
+
+    player1Array.push(player1SecondRollOutput);
+    console.log("pushed player1secondrolloutput into player1array");
```
lim-jiahao, line 71: From the rest of your code, you don't use the `player1FirstRollOutput` and `player1SecondRollOutput` variable anywhere else. Seems like its only used here as a storage variable before you push it to `player1Array`. Consider how you can simplify this, there's multiple ways to do so but one way I would do it that reduces all this to 1 line is simply `player1Array = [diceRoll(), diceRoll()]` This way, you also dont need the 4 variables declared in lines 32 to 35 which cleans up the code

script.js
```diff
@@ -1,4 +1,210 @@
...
+
+    console.log(player1Array);
+
+    currentGameMode = player1Select; //change game mode as soon as dices are rolled so it won't re-roll
+    console.log("game mode is changed to player1select");
+
+    myOutputValue = `<b>WELCOME PLAYER 1 </b>🎲🎲🎲 <br><br>You rolled ${player1FirstRollOutput} for Dice 1 and ${player1SecondRollOutput} for Dice 2.<br><br>${selectDiceOrderText} `;
+    return myOutputValue;
```
lim-jiahao, line 79: Again, try not to get too used to the use of `myOutputValue`, can be renamed to other more descriptive names if you want or as mentioned above, you can just return the string directly here

script.js
```diff
@@ -1,4 +1,210 @@
...
+  }
+
+  //game mode: player 1 to select which dice first
+  if (currentGameMode == player1Select) {
+    // how to collect user choice of which dice goes first? (kiv)
+    userInput = Number(input);
```
lim-jiahao, line 85: This is not a global variable and its the first time its used in the function so do remember to declare it with `var`

script.js
```diff
@@ -1,4 +1,210 @@
...
+    //for (1) nan inputs
+    if (Number.isNaN(userInput) == true) {
+      myOutputValue = "sorry, please enter a number.";
+      return myOutputValue;
+    }
+
+    if (userInput != 1 && userInput != 2) {
+      myOutputValue = "Please insert only either '1' or '2'.";
+      return myOutputValue;
+    }
```
lim-jiahao, line 99: Good input validation again, though I personally would combine the two. Reason being, even if the user enters something that is not a number, you want to tell them to enter either a 1 or 2 anyway, not just any number. So this can be reduced to simply

```
if (userInput != 1 && userInput != 2) {
    return "Please insert only either '1' or '2'."
}
```
The NaN scenario is already taken care of by this, because anything that is not a number will not be equal to 1 and not be equal to 2, thus evaluating to true and coming to this condition

Also, one thing you might wanna consider is showing the dice rolls again in this message if they don't enter a valid 1 or 2. Because let's say I did not enter a 1 or 2 and end up getting this message, I might have forgotten already what my dice rolls are and need a reminder. Just another possible enhancement of user experience

script.js
```diff
@@ -1,4 +1,210 @@
...
+      console.log("game mode is changed to player2roll");
+      return myOutputValue;
+    } else if (userInput == 2) {
+      player1Number = String(player1Array[1]) + String(player1Array[0]);
+      myOutputValue = `Player 1, you chose Dice ${userInput} first.<br><br>You rolled ${player1Array}.<br><b>Your number is ${player1Number}.</b><br><br>${player2TurnText} `;
+      currentGameMode = player2Roll; // change game mode should be within conditionals
+      console.log("game mode is changed to player2roll");
+
+      return myOutputValue;
+    }
```
lim-jiahao, line 116: Good way to calculate the score of the user. But notice how your lines 105-108 and lines 111-115 are repeated and doing the same thing? In programming, we try not to repeat ourselves if possible - the reason for this is 1. for readability purpose, and 2. for code maintainability purpose, imagine you notice a bug here and need to change something, then you need to change it twice for the code sections you repeated. And this gets more complex if there are more instances of unnecessary repetition in your code. 

Consider having an if-else that calculates the `player1Number` based on the desired order, then doing the other stuff after the if-else.

```
    if (userInput == 1) {
      //if input is 1, let the first number go first. Else if input is 2, let second number go first
      player1Number = String(player1Array[0]) + String(player1Array[1]);
    } else if (userInput == 2) {
      player1Number = String(player1Array[1]) + String(player1Array[0]);
    }

      myOutputValue = `Player 1, you chose Dice ${userInput} first.<br><br>You rolled ${player1Array}.<br><b>Your number is ${player1Number}.</b><br><br>${player2TurnText}`;
      currentGameMode = player2Roll; // change game mode should be within conditionals
      console.log("game mode is changed to player2roll");
      return myOutputValue;
```

script.js
```diff
@@ -1,4 +1,210 @@
...
+    myOutputValue = `<b>WELCOME PLAYER 2 </b>🎲🎲🎲<br><br>You rolled ${player2FirstRollOutput} for Dice 1 and ${player2SecondRollOutput} for Dice 2.<br><br>${selectDiceOrderText}`;
+    return myOutputValue;
+  }
+  // game mode: player 2 to select
+
+  userInput = Number(input);
+  var textToProceedToCheckWinner = "Click submit to find out who won!🧐🤔";
+
+  if (currentGameMode == player2Select) {
+    //input validation for (1) not a number (2) not 1 and not 2
```
lim-jiahao, line 137: the code for this game mode and the previous (player2Roll) is almost entirely similar to the code for game modes `player1Roll` and `player1Select`, other than just the change in player and the addition of your `textToProceedToCheckWinner`. consider how you can abstract these into its own functions, maybe say `playerRoll` and `playerSelect`. You can then have a global variable called `curPlayer` which tracks who the current player is. This way you can customise your output and the array you calculate the score from based on this global variable.

This one is a bit complicated for me to show an example but do give it a shot and let me know if you face issues! This would clean up your code significantly by 1. reducing repetition and 2. separating different aspects of the gameplay into its own function which makes your `beatThatGame` function shorter. One thing you wanted specific feedback on is how to refactor code so this would definitely be an area that I would start with!

script.js
```diff
@@ -1,4 +1,210 @@
...
+      isTheGameOver = true; //game is over, hence game is over
+
+      return myOutputValue;
+    } else if (Number(player1Number) < Number(player2Number)) {
+      myOutputValue = `${outputText}<br><br>Player 2 got a larger number than player 1. 
+    <br><br><b>Player 2 wins!</b> 🎉🔥🥳<br><br> Click submit to play again.`;
+      isTheGameOver = true;
+      return myOutputValue;
+    }
+  }
```
lim-jiahao, line 188: This is ok, but since you're returning early, you don't need the else ifs and the last condition. 

```
    if (Number(player1Number) == Number(player2Number)) {
      myOutputValue = `${outputText}<br>You've gotten a draw. Please try again.`;

      return myOutputValue;
    }

   if (Number(player1Number) > Number(player2Number)) {
      myOutputValue = `${outputText}<br><br>Player 1 got a larger number than Player2. <br><br><b>Player 1 wins!</b> 🎉🔥🥳<br><br> Click submit to play again.`;
      isTheGameOver = true; //game is over, hence game is over

      return myOutputValue;
    } 

    myOutputValue = `${outputText}<br><br>Player 2 got a larger number than player 1. 
  <br><br><b>Player 2 wins!</b> 🎉🔥🥳<br><br> Click submit to play again.`;
    isTheGameOver = true;
    return myOutputValue;
```

The way the code above works is: if the first condition is met return early, otherwise it carries on to the second condition (which is why we don't need the else if), again if it is met then we return early. Otherwise we carry on to the last part which does not need to be in an `if` statement, this serves as our 'default' case if we didn't return early

script.js
```diff
@@ -1,4 +1,210 @@
...
+
+//crete function that resets game mode by reassigning game mode, array, and reset isTheGameFalse
+var resetGame = function () {
+  //reset game
+  currentGameMode = player1Roll;
+  console.log("the game is reset");
+  player1Array = [];
+  player2Array = [];
+  isTheGameOver = false;
+};
```
lim-jiahao, line 199: Nice use of a resetGame function!

script.js
```diff
@@ -1,4 +1,210 @@
...
+
+//main function
 var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
+  var finalOutput = beatThatGame(input);
```
lim-jiahao, line 203: I discussed how you can do this in the above comments, but `beatThatGame` is a huge function so do consider how you can refactor it. Your game and code can be broken into three logical parts: 1. players roll, 2. players select order, 3. compare score. This provides a very nice starting point on how we can split `beatThatGame` further into separate functions.

lim-jiahao:
Hey Raymond,

Nice work and code is generally very clean and readable! You have a good grasp of all the concepts taught so far and it really showed in your project. You are right in feeling pleased about the user-friendliness of your game -  when loading it up to play it, at no point was I lost on what to do so that's definitely a plus point! Most developers just focus on the coding aspect of things without considering that at the end of the day, we are designing for users. 

For your desired specific feedback, I've already given some pointers on how the refactoring can be done to make the code look more organised and cleaner in my comments. Do look through them and feel free to look for me if you have any questions. 

Hope this helps and I can't wait to see what you come up with for blackjack! Keep it up :)

script.js
lim-jiahao: also consider a case where the user rolled the same number for both die, does it make sense from a game and user pov to ask them for their desired order? Multiple ways to incorporate this!

PR #285, opened 2021-12-11
B12-2 Xian Kai Beat That! Submission

xiankaiheng:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

3.5 hours

Issue with Syntax

To find out how to change modes smoothly between each player and turns

Comfort Level (1-5):
3

Completeness Level (1-5):
3

It is one that tests quite a bit on workflow logic which is useful.

-

How to better visualise nested ifs.

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,5 @@
 /*Base
+clg clg 
 Requirements
 There are 2 players and players take turns.
 When a player clicks Submit, the game rolls 2 dice and shows the dice rolls, for example 3 and 6.
--- a/script.js
+++ b/script.js
@@ -1,4 +1,184 @@
+/*Base
+Requirements
+There are 2 players and players take turns.
+When a player clicks Submit, the game rolls 2 dice and shows the dice rolls, for example 3 and 6.
+The player picks the order of the dice they want. For example, if they wanted the number 63, they would specify that the 2nd dice goes first. You can choose how the player specifies dice order.
+After both players have rolled and chosen dice order, the player with the higher combined number wins.*/
+
+// -1: Introduction and game mode selection
+// 0 : Before player 1 roll dice
+// 1 : After player 1 roll dice
+// 2 : Before player 2 roll dice
+// 3 : After player 2 roll dice
+var turnOrOrder = -1;
+
+// 1: Normal mode
+// 2: Reversed Mode
+var gameMode = 0;
+
+// 0 : No action needed
+// 1 : Action needed
+var whichDiceOrder = 0;
+
+// to track overall results
+var pOneStats = 0;
+var pTwoStats = 0;
+var leaderboard = "";
+var myOutputValue;
+
+var playerThrow = [0, 0];
+var pOneResult = 0;
+var pTwoResult = 0;
+
+// Dice roll function to get random dice number
+var diceRollResult = function () {
+  var rollResult = Math.floor(Math.random() * 6) + 1;
+  return rollResult;
+};
+
+// Evaluate results
+var evaluateResults = function (input) {
+  //For both player first roll
+  if ((turnOrOrder == 0 || turnOrOrder == 2) && whichDiceOrder == 0) {
+    whichDiceOrder = 1;
+    turnOrOrder += 1;
+    playerThrow[0] = diceRollResult();
+    playerThrow[1] = diceRollResult();
+    myOutputValue = `You rolled ${playerThrow[0]} and ${playerThrow[1]}.`;
+    console.log(turnOrOrder);
+
+    //If both die result is the same, skip order selection
+    if (playerThrow[0] == playerThrow[1]) {
+      whichDiceOrder = 0;
+      turnOrOrder += 1;
+      //result for P1
+      if (turnOrOrder == 2) {
+        pOneResult = "" + playerThrow[0] + playerThrow[1];
+        //result for P2
+        myOutputValue =
+          myOutputValue +
+          `<br><br>Your hand is ${playerThrow[0]}${playerThrow[1]}.`;
+        return myOutputValue;
+      } else if (turnOrOrder == 4) {
+        pTwoResult = "" + playerThrow[0] + playerThrow[1];
+        myOutputValue =
+          myOutputValue +
+          `<br><br>Your hand is ${playerThrow[0]}${playerThrow[1]}.`;
+        return myOutputValue;
+      }
+    }
+
+    //Else ask player which order they want to choose
+    myOutputValue =
+      myOutputValue +
+      ` <br><br>Which order would you like to put them?<br> 1) die 1 first : ${playerThrow[0]}${playerThrow[1]} <br> 2) die 2 first : ${playerThrow[1]}${playerThrow[0]}`;
+    return myOutputValue;
+  } // Player to choose order of die:
+  else if (turnOrOrder == 1 && whichDiceOrder == 1) {
+    // Player 1 return message (die 1 first)
+    if (input == 1) {
+      turnOrOrder += 1;
+      whichDiceOrder = 0;
+      pOneResult = "" + playerThrow[0] + playerThrow[1];
+      return `Ok, your hand is ${pOneResult}. <br><br>It is now Player 2's turn.`;
+    } // player 1 return message (die 2 first)
+    else if (input == 2) {
+      turnOrOrder += 1;
+      whichDiceOrder = 0;
+      pOneResult = "" + playerThrow[1] + playerThrow[0];
+      return `Ok, your hand is ${pOneResult}. <br><br>It is now Player 2's turn.`;
+    } else {
+      return `Please select a valid number.<br><br>${myOutputValue}`;
+    }
+  } // Player 2 choose order of die
+  else if (turnOrOrder == 3 && whichDiceOrder == 1) {
+    if (input == 1) {
+      console.log("p2 1");
+      turnOrOrder = 4;
+      whichDiceOrder = 0;
+      pTwoResult = "" + playerThrow[0] + playerThrow[1];
+      myOutputValue = `Ok, your hand is ${pTwoResult}.`;
+      return myOutputValue;
+    } else if (input == 2) {
+      console.log("p2 2");
+      turnOrOrder = 4;
+      whichDiceOrder = 0;
+      pTwoResult = "" + playerThrow[1] + playerThrow[0];
+      myOutputValue = `Ok, your hand is ${pTwoResult}`;
+      return myOutputValue;
+    } else {
+      return `Please select a valid number.<br><br>${myOutputValue}`;
+    }
+  }
+
+  // Evaluate results for winner
+  else if (turnOrOrder == 4 && whichDiceOrder == 0) {
+    turnOrOrder = 0;
+    pOneStats = Math.floor(pOneStats) + Math.floor(pOneResult);
+    pTwoStats = Math.floor(pTwoStats) + Math.floor(pTwoResult);
+    gameStats();
+    //P2 results bigger
+    // add in game mode condition
+    if (
+      (pOneResult > pTwoResult && gameMode == 1) ||
+      (pOneResult < pTwoResult && gameMode == 2)
+    ) {
+      myOutputValue =
+        myOutputValue +
+        ` but Player 1's hand is ${pOneResult}. You lost :(<br><br>The game will now restart!` +
+        leaderboard;
+      return myOutputValue;
+      //P2 results smaller
+    } else if (
+      (pTwoResult > pOneResult && gameMode == 1) ||
+      (pTwoResult < pOneResult && gameMode == 2)
+    ) {
+      myOutputValue =
+        myOutputValue +
+        ` and Player 1's hand is ${pOneResult}. You won! :)<br><br>The game will now restart!` +
+        leaderboard;
+      return myOutputValue;
+      //P2 results smame
+    } else if (pTwoResult == pOneResult) {
+      myOutputValue =
+        myOutputValue +
+        ` and Player 1's hand is ${pOneResult}. It's a draw! :)<br><br>The game will now restart!` +
+        leaderboard;
+      return myOutputValue;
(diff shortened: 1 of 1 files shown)


script.js
```diff
@@ -1,4 +1,184 @@
...
+// 3 : After player 2 roll dice
+var turnOrOrder = -1;
+
+// 1: Normal mode
+// 2: Reversed Mode
+var gameMode = 0;
+
+// 0 : No action needed
+// 1 : Action needed
+var whichDiceOrder = 0;
```
lim-jiahao, line 22: Good use of these three variables and numbers to store state information about your game, but truth be told it was quite hard to follow the code when reading through and I had to repeatedly refer back here to know what each number means. 

For `gameMode` and `whichDiceOrder`, since they can only take two possible values, consider using a boolean for them. So let's say `gameMode`, it can be called `isReverseMode` and set to be `false` at first. Then subsequently in your code where you check say, `if (gameMode == 2)` you can do `if (isReverseMode)` and similarly for normal mode. You can do the same for `whichDiceOrder`, a better name would prob be `isActionNeeded` which is also set to `false` at first. 

This makes your code more readable - the variable name is clearer and the use of the boolean variable means that the code can read more like English. This helps especially when working in large teams where many developers work together. Ensuring that your code is clean and readable for not just for yourself but for your teammates is vital. 

For `turnOrOrder` I get your intention to just have this one variable track all different possible turn states. But considering that this is the key game state which your whole gameplay logic hinges on, I would sacrifice succinctness here for more explicit and elaborate turn states, which again provides the benefit of more readable code in your gameplay below.

```
const GAME_MODE_SELECTION = -1;
const BEFORE_PLAYER_1_ROLL = 0;
const AFTER_PLAYER_1_ROLL = 1;
const BEFORE_PLAYER_2_ROLL = 2;
const AFTER_PLAYER_2_ROLL = 3;
var turnOrOrder = GAME_MODE_SELECTION;
```
(Using the `const` keyword and SCREAM_CASE here to further emphasise its a constant, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const for more info. You will learn this in bootcamp too so good to get some exposure early)

Then in your code, you set turnOrOrder to the relevant variable as the game progresses. Then for example in line 40 when you are checking `if ((turnOrOrder == 0 || ...` it can be `turnOrOrder = BEFORE_PLAYER_1_ROLL` which is much more descriptive. I can instantly know what turn state is being checked for or being updated to without having to reference back up here.

script.js
```diff
@@ -1,4 +1,184 @@
...
+var pTwoResult = 0;
+
+// Dice roll function to get random dice number
+var diceRollResult = function () {
+  var rollResult = Math.floor(Math.random() * 6) + 1;
+  return rollResult;
+};
+
+// Evaluate results
+var evaluateResults = function (input) {
```
lim-jiahao, line 41: `evaluateResults` is a huge function so do consider how you can refactor it. Your gameplay logic and code can be broken into three distinct parts: 1. players roll, 2. players select order, 3. compare score. This provides a very nice starting point on how we can split `evaluateResults` further into separate functions.

script.js
```diff
@@ -1,4 +1,184 @@
...
+  //For both player first roll
+  if ((turnOrOrder == 0 || turnOrOrder == 2) && whichDiceOrder == 0) {
+    whichDiceOrder = 1;
+    turnOrOrder += 1;
```
lim-jiahao, line 45: this is an example of what i meant earlier when i mentioned that its hard to keep track, right here and subsequently for the rest of your code, i have to commit to memory your comment up top on what each number means. but if here i forget what 0 1 or 2 for `turnOrder` means or what 0 or 1 for `whichDiceOrder` means, its hard to follow the rest of your code

script.js
```diff
@@ -1,4 +1,184 @@
...
+    playerThrow[0] = diceRollResult();
+    playerThrow[1] = diceRollResult();
```
lim-jiahao, line 47: this can just be `playerThrow = [diceRollResult(), diceRollResult()]`
for larger amount of dice rolls, do consider a loop!

script.js
```diff
@@ -1,4 +1,184 @@
...
+          `<br><br>Your hand is ${playerThrow[0]}${playerThrow[1]}.`;
+        return myOutputValue;
+      } else if (turnOrOrder == 4) {
+        pTwoResult = "" + playerThrow[0] + playerThrow[1];
+        myOutputValue =
+          myOutputValue +
+          `<br><br>Your hand is ${playerThrow[0]}${playerThrow[1]}.`;
+        return myOutputValue;
+      }
+    }
```
lim-jiahao, line 70: Lines 58-61 and lines 64-67 are repetitive and do the same thing. In programming, we try not to repeat ourselves if possible - the reason for this is 1. for readability purpose, and 2. for code maintainability purpose, imagine you notice a bug here and need to change something, then you need to change it twice for the code sections you repeated. And this gets more complex if there are more instances of unnecessary repetition in your code.

Consider having an if-else that stores the result in either pOneResult or pTwoResult based on the value of turnOrOrder, then doing the other stuff after the if-else.

An example for this would be:
```
    if (playerThrow[0] == playerThrow[1]) {
      whichDiceOrder = 0;
      turnOrOrder += 1;
      //result for P1
      if (turnOrOrder == 2) {
        pOneResult = "" + playerThrow[0] + playerThrow[1];
      } else if (turnOrOrder == 4) {
        pTwoResult = "" + playerThrow[0] + playerThrow[1];
      }
        myOutputValue =
          myOutputValue +
          `<br><br>Your hand is ${playerThrow[0]}${playerThrow[1]}.`;
        return myOutputValue;
    }
```

script.js
```diff
@@ -1,4 +1,184 @@
...
+      return `Ok, your hand is ${pOneResult}. <br><br>It is now Player 2's turn.`;
+    } // player 1 return message (die 2 first)
+    else if (input == 2) {
+      turnOrOrder += 1;
+      whichDiceOrder = 0;
+      pOneResult = "" + playerThrow[1] + playerThrow[0];
+      return `Ok, your hand is ${pOneResult}. <br><br>It is now Player 2's turn.`;
+    } else {
+      return `Please select a valid number.<br><br>${myOutputValue}`;
+    }
```
lim-jiahao, line 93: commenting this here but this actually applies to the bulk of your code. you mentioned that a specific feedback you want is how to visualise nested `if`s. i did notice that your code has many layers of nesting and one thing I emphasised since the start is making use of early return, which you are doing. but when we return early there is usually no need to nest the rest of your code in additional `else if`, `else`, etc. 

An example for this little section here:
```
    if (input == 1) {
      turnOrOrder += 1;
      whichDiceOrder = 0;
      pOneResult = "" + playerThrow[0] + playerThrow[1];
      return `Ok, your hand is ${pOneResult}. <br><br>It is now Player 2's turn.`;
    } // player 1 return message (die 2 first)
    
    if (input == 2) {
      turnOrOrder += 1;
      whichDiceOrder = 0;
      pOneResult = "" + playerThrow[1] + playerThrow[0];
      return `Ok, your hand is ${pOneResult}. <br><br>It is now Player 2's turn.`;
    } 

    return `Please select a valid number.<br><br>${myOutputValue}`;
```
See how the `else if` and `else` is gone altogether? The way the code above works is: if the first condition is met return early, otherwise it carries on to the second condition (which is why we don't need the else if), again if it is met then we return early. Otherwise we carry on to the last part which does not need to be in a conditional statement, this serves as our 'default' case if we didn't return early.

By applying this fix to other parts of your code, I'm sure it will reduce your nested `if`s drastically which will eliminate the problem of having trouble visualising the layers of nesting.

script.js
```diff
@@ -1,4 +1,184 @@
...
+    } else if (input == 2) {
+      console.log("p2 2");
+      turnOrOrder = 4;
+      whichDiceOrder = 0;
+      pTwoResult = "" + playerThrow[1] + playerThrow[0];
+      myOutputValue = `Ok, your hand is ${pTwoResult}`;
+      return myOutputValue;
+    } else {
+      return `Please select a valid number.<br><br>${myOutputValue}`;
+    }
```
lim-jiahao, line 112: same as above, we are making use of early return so no need to nest if / else if / else

script.js
```diff
@@ -1,4 +1,184 @@
...
+  }
+
+  // Evaluate results for winner
+  else if (turnOrOrder == 4 && whichDiceOrder == 0) {
+    turnOrOrder = 0;
+    pOneStats = Math.floor(pOneStats) + Math.floor(pOneResult);
+    pTwoStats = Math.floor(pTwoStats) + Math.floor(pTwoResult);
```
lim-jiahao, line 119: aren't the results always integers? why the need for the Math.floor

script.js
```diff
@@ -1,4 +1,184 @@
...
+    gameStats();
```
lim-jiahao, line 120: Function name could be more descriptive, the recommendation is to always begin function names with a verb that clearly describes what it does. In this case, it seems like `gameStats` is updating the leaderboard so `updateLeaderboard` might be better. This way I don't even need to look at the function definition to know what it does, whereas `gameStats` is a bit vague

script.js
```diff
@@ -1,4 +1,184 @@
...
+    return `Welcome players, lets play a game of Beat That! <br><br>Rules:<br>Each player will take turn to roll 2 dice and choose the order they want to arrange the dice. <br><br> In normal mode, the player with the bigger hand wins while in reversed mode, the player with the smaller hand wins!<br><br> Please indicate whether you would like to play normal or reversed mode:<br>1) Normal Mode<br>2) Reversed Mode`;
+  } else if (turnOrOrder == -1 && (input == 1 || input == 2)) {
+    turnOrOrder += 1;
+    if (input == 1) {
+      gameMode = 1;
+      return `All right, lets start with the Normal game mode! Player 1 please click to roll.`;
+    } else {
+      gameMode = 2;
+      return `All right, lets start with the Reversed game mode! Player 1 please click to roll.`;
+    }
```
lim-jiahao, line 160: I would put this at the top of the function, always cater for the special or rarer scenarios first and return early, then move on to the rest of your code. This also makes sense intuitively because your code actually starts out with `turnOrOrder` being -1 and comes in here first. 

Also, try seeing how if you can put line 150 into the output box immediately on page load by editing the html so that the user sees the message straight away when he loads the game. Currently when I load your game, I am greeted with an empty text input and output box, and it is not immediately clear what I have to do. Always think about user experience when you are designing something!

Also for the case in line 151 onwards, some input validation is probably needed because if I enter something that is not 1 or 2, I get back 'undefined' in the output box.

script.js
lim-jiahao: i would also put the negative case up top first and check for that

```
    if (input != 1 && input != 2) {
      return `Please select a valid number.<br><br>${myOutputValue}`;
    }

    turnOrOrder = 4;
    whichDiceOrder = 0;

    if (input == 1) {
      console.log("p2 1");
      pTwoResult = "" + playerThrow[0] + playerThrow[1];
      myOutputValue = `Ok, your hand is ${pTwoResult}.`;
      return myOutputValue;
    } 
    
    console.log("p2 2");
    pTwoResult = "" + playerThrow[1] + playerThrow[0];
    myOutputValue = `Ok, your hand is ${pTwoResult}`;
    return myOutputValue;
```

lim-jiahao: but good input validation here and i like that you display the 'please select a valid number' message to the user along with the dice rolls again! a comment i have for many students is that when they output this msg, they don't show the rolls again and it is very possible the user forgot the dice rolls

script.js
```diff
@@ -1,4 +1,184 @@
...
+  } else if (pOneStats == pTwoStats) {
+    leaderboard = `<br><br><br>It's a draw!<br><br> Player 1 Score: ${pOneStats}<br>Player 2 Score: ${pTwoStats} `;
+  }
+};
+
 var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
+  return evaluateResults(input);
 };
```
lim-jiahao, line 185: this `main` function into `evaluateResults` is quite redundant if all that `main` does is taking in `input` and returning the result of calling `evaluateResults` with `input` so again, consider how you can refactor `evaluateResults` into further smaller functions and call them in `main`.

lim-jiahao:
Hey Xian Kai,

Good job and I like that you tried out reverse mode too! You have very strong grasp of the concepts taught so far and put it to good use here. Hope it was just a formatting issue in the questionnaire and that's why you don't have an answer to what you're pleased about. I think you have a lot to be pleased about here, especially to come up with something like this in just 3.5 hrs. My comments are admittedly a bit harsher but it was on purpose as I feel they will help you out a lot especially as you move into Bootcamp. 

High level comments: (which I expanded on in my other comments)
1. more descriptive variable names, and better tracking of state
2. refactoring of evaluateResults function into smaller functions -> maybe 1 for rolling die, 1 for choosing die order, 1 for comparing score
3. reducing the nesting of your if / else ifs by making use of the fact that you return early

Also, I realise you didn't have a chance to in this project but do try playing around with HTML/CSS for blackjack! The first part of bootcamp is mainly front-end doing DOM manipulation using HTML/CSS and JavaScript so it would be good to have some exposure right now, even if it is simple things like changing the background color of the page or the container box, or changing the fonts that are being used. 

Very eager to see what you will come up with for blackjack!

script.js
```diff
@@ -1,4 +1,184 @@
...
+    gameStats();
+    //P2 results bigger
+    // add in game mode condition
+    if (
+      (pOneResult > pTwoResult && gameMode == 1) ||
+      (pOneResult < pTwoResult && gameMode == 2)
+    ) {
+      myOutputValue =
+        myOutputValue +
+        ` but Player 1's hand is ${pOneResult}. You lost :(<br><br>The game will now restart!` +
```
lim-jiahao, line 129: seems like the output on who wins/loses is always from the pov of player 2, was this design choice intentional? small issue regardless, but personal preference would be for the output to say specifically which player won

PR #286, opened 2021-12-11
Beat That (Base) Submission (Section 12-2 Han Chun)

hanchunz:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

4 hours

Initially, I did not have any input validation and if the players did not follow the flow of the game, the output box will simply show undefined. Thus, I spent quite some time making the input validation to ensure there will always be a prompt in the output box to guide the players to continue the game from where they left off.

Reviewing the code and refactoring it.

Comfort Level (1-5): 4

4

Completeness Level (1-5):

3

It was a good practice in applying what we have learnt so far. It allows us to use the latest lesson materials (arrays and loops) to help make our code shorter and more concise, even though we could have actually completed the project with the earlier lessons only but with longer code

Being able to comment it out in a nice manner so anyone can get a sense of the project.

No specific aspects, though any point of improvement would be much appreciated. :)

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,275 @@
+//Create a version of the Beat That dice game, where players create the largest number they can by concatenating random dice roll digits.
+
+/* 
+Rules:
+
+1) There are 2 players and players take turns.
+
+2) When a player clicks Submit, the game rolls 2 dice and shows the dice rolls, for example 3 and 6.
+
+3) The player picks the order of the dice they want. For example, if they wanted the number 63, they would specify that the 2nd dice goes first. You can choose how the player specifies dice order.
+
+4) After both players have rolled and chosen dice order, the player with the higher combined number wins.
+*/
+
+//Create global variables that stores whether the players have rolled
+var hasPlayer1Rolled = 0;
+var hasPlayer2Rolled = 0;
+
+//Create a global variable that is an array that will store the players' rolled numbers. It is empty at first but when the numbers are pushed in it will be in the following format ideally: [player1Dice1, player1Dice2] [player2Dice1, player2Dice2]
+var player1DiceRolls = [];
+var player2DiceRolls = [];
+
+//Create global variables for the players' concatenated numbers
+var player1ConcatenatedNumber = 0;
+var player2ConcatenatedNumber = 0;
+
+//Create a global variable to store the reset mode
+var timeToReset = 0;
+
+//Create a function that capitalizes the first letter of a string
+var capitalizer = function (userInput) {
+  return userInput[0].toUpperCase() + userInput.substring(1);
+};
+
+//Create the dice roll function
+var randomDiceNumberGenerator = function () {
+  return Math.floor(Math.random() * 6) + 1;
+};
+
+//Create a function to roll the players' dices, depending on whether its player 1 turn or player 2 turn
+var rollTheDice = function () {
+  //If Player 1 has not rolled (hasPlayer1Rolled = 0), roll the dice twice and push the numbers to the player1DiceRolls array. If Player 1 has rolled (hasPlayer1Rolled = 1), then check if player 2 has not rolled (hasPlayer2Rolled = 0), roll the dice twice and push the numbers to the player2DiceRolls array.
+  if (hasPlayer1Rolled == 0) {
+    //Push two dice rolls into Player 1 Dice Roll's array, player1DiceRolls, using a for loop
+    for (var counter = 0; counter < 2; counter += 1) {
+      player1DiceRolls.push(randomDiceNumberGenerator());
+    }
+    console.log(
+      "Player 1 Dice 1: ",
+      player1DiceRolls[0],
+      "Player 1 Dice 2: ",
+      player1DiceRolls[1]
+    );
+
+    //Change the global variable hasPlayer1Rolled to 1 to show that Player 1 has rolled
+    hasPlayer1Rolled = 1;
+  } else if (hasPlayer1Rolled == 1 && hasPlayer2Rolled == 0) {
+    //Push two dice rolls into Player 2 Dice Roll's array, player1DiceRolls, using a for loop
+    for (var counter = 0; counter < 2; counter += 1) {
+      player2DiceRolls.push(randomDiceNumberGenerator());
+    }
+    console.log(
+      "Player 2 Dice 1: ",
+      player2DiceRolls[0],
+      "Player 2 Dice 2: ",
+      player2DiceRolls[1]
+    );
+    //Change the global variable hasPlayer1Rolled to 1 to show that Player 1 has rolled
+    hasPlayer2Rolled = 1;
+  }
+};
+
+//Create a function that concatenates the players' two dice numbers
+var concatenator = function (userInput) {
+  //Check if player 1 has rolled and player 2 has not
+  if (hasPlayer1Rolled == 1 && hasPlayer2Rolled == 0) {
+    //Check if player 1 types in dice 1 or dice 2
+    if (userInput == "dice 1") {
+      player1ConcatenatedNumber = Number(
+        String(player1DiceRolls[0]) + String(player1DiceRolls[1])
+      );
+      console.log(
+        "Player 1's concatenated number is: ",
+        player1ConcatenatedNumber
+      );
+    } else if (userInput == "dice 2") {
+      player1ConcatenatedNumber = Number(
+        String(player1DiceRolls[1]) + String(player1DiceRolls[0])
+      );
+      console.log(
+        "Player 1's concatenated number is: ",
+        player1ConcatenatedNumber
+      );
+    }
+  }
+
+  //Check if player 1 has rolled and player 2 has rolled
+  if (hasPlayer1Rolled == 1 && hasPlayer2Rolled == 1) {
+    //Check if player 2 types in dice 1 or dice 2
+    if (userInput == "dice 1") {
+      player2ConcatenatedNumber = Number(
+        String(player2DiceRolls[0]) + String(player2DiceRolls[1])
+      );
+      console.log(
+        "Player 2's concatenated number is: ",
+        player2ConcatenatedNumber
+      );
+    } else if (userInput == "dice 2") {
+      player2ConcatenatedNumber = Number(
+        String(player2DiceRolls[1]) + String(player2DiceRolls[0])
+      );
+      console.log(
+        "Player 2's concatenated number is: ",
+        player2ConcatenatedNumber
+      );
+    }
+  }
+};
+
+//Create a function that outputs the notification statement after the players' have rolled the dice on what their dice number is.
+var statementForDiceRolled = function () {
+  //Check if only player 1 has rolled
+  if (hasPlayer1Rolled == 1 && hasPlayer2Rolled == 0) {
+    return `Welcome Player 1. <br><br>You rolled ${player1DiceRolls[0]} for Dice 1 and ${player1DiceRolls[1]} for Dice 2. <br><br>Choose the order of the dice. <br>Input "dice 1" if you wish for Dice 1's number to be in front of Dice 2's number. <br>Input "dice 2" if you wish for Dice 2's number to be in front of Dice 1's number.`;
+  }
+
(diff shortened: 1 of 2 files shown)


script.js
```diff
@@ -1,4 +1,275 @@
...
+//Create a global variable that is an array that will store the players' rolled numbers. It is empty at first but when the numbers are pushed in it will be in the following format ideally: [player1Dice1, player1Dice2] [player2Dice1, player2Dice2]
+var player1DiceRolls = [];
+var player2DiceRolls = [];
+
+//Create global variables for the players' concatenated numbers
+var player1ConcatenatedNumber = 0;
+var player2ConcatenatedNumber = 0;
+
+//Create a global variable to store the reset mode
+var timeToReset = 0;
```
lim-jiahao, line 28: Nice and clear global variables. For `hasPlayer1Rolled`, `hasPlayer2Rolled` and `timeToReset`, since they can only take values 0 and 1, you are implicitly using a boolean variable (more formally, 0 is falsy and 1 is truthy, see [here](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) and [here](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) for more details). my suggestion would just be to make them boolean variables then. so they will both be assigned `false` first, and updated to true in the relevant places in your code

script.js
```diff
@@ -1,4 +1,275 @@
...
+
+//Create a function that capitalizes the first letter of a string
+var capitalizer = function (userInput) {
```
lim-jiahao, line 31: nice use of this helper function to standardise the user input regardless of case, good thinking on the user experience!

script.js
```diff
@@ -1,4 +1,275 @@
...
+  return userInput[0].toUpperCase() + userInput.substring(1);
+};
+
```
lim-jiahao, line 34: nice use of this helper function to standardise the user input regardless of case, good thinking on the user experience!

script.js
```diff
@@ -1,4 +1,275 @@
...
+  player1ConcatenatedNumber = 0;
+  player2ConcatenatedNumber = 0;
+  timeToReset = 0;
+  return `The game has been reset! You may play again by pressing the "Submit" button.`;
+};
+
+//Create a function that receives the user input and returns the correct output statements
+var beatThatGame = function (userInput) {
+  //Make the input entirely lowercase
+  userInput = userInput.toLowerCase();
```
lim-jiahao, line 186: good standardising of user input regardless of case, nice that you're thinking about the user experience!

script.js
```diff
@@ -1,4 +1,275 @@
...
+    return gameReset();
+  }
+  console.log("Player 1: ", player1DiceRolls, "Player 2: ", player2DiceRolls);
+};
+
+//Create a function that takes the user input and outputs the correct instructions if the user types invalid input
+var invalidInputChecker = function () {
+  //Check if both players have not rolled the dice yet
+  if (hasPlayer1Rolled == 0 && hasPlayer2Rolled == 0) {
+    return `Please press the "submit" button without typing anything in the input box to start the game for player 1`;
```
lim-jiahao, line 230: Do we really need this specific condition where the input box needs to be empty in order to start the game? Seems oddly specific, why not just ignore whatever input and start the game regardless once the submit button is clicked, if both players have not rolled any die

To build on this further, would be a good idea to output this message to the user immediately on page load, instructing them how to start the game. This can be done by changing your HTML, do explore and let me know if you have questions. Currently, when I load up your game, I am greeted with an empty text input and empty output box, and it is not immediately clear what I have to do. Another point related to user experience to consider

script.js
```diff
@@ -1,4 +1,275 @@
...
+
+  if (
+    !(userInput == "dice 1" || userInput == "dice 2") &&
+    ((hasPlayer1Rolled == 1 && player1ConcatenatedNumber == 0) ||
+      (hasPlayer1Rolled == 1 &&
+        hasPlayer2Rolled &&
+        player2ConcatenatedNumber == 0))
+  ) {
+    return;
+  }
```
lim-jiahao, line 196: this is definitely a unique way to do input validation by returning undefined here and subsequently calling your `invalidInputChecker`. I have mixed opinions about this and can see the benefits of doing it this way but one point I have is that the conditions in the `if` statement here are very complicated and hard to decipher. 

I would much rather incorporate your input validation into this `beatThatGame` function and do an individual check for each invalid input scenario and return the relevant output msg.

script.js
```diff
@@ -1,4 +1,275 @@
...
+    console.log(
+      "Player 2 Dice 1: ",
+      player2DiceRolls[0],
+      "Player 2 Dice 2: ",
+      player2DiceRolls[1]
+    );
+    //Change the global variable hasPlayer1Rolled to 1 to show that Player 1 has rolled
+    hasPlayer2Rolled = 1;
+  }
+};
```
lim-jiahao, line 71: this function is ok and nice use of loops to push your dice rolls in. since we are only dealing with 2 rolls right now, maybe a loop is unnecessary and you can just do something like `player1DiceRolls = [randomDiceNumberGenerator(), randomDiceNumberGenerator()]` but definitely use loops if we are dealing with more dice rolls! 

Feels like this function can be merged together with `statementForDiceRolled` where the return statements are incorporated into `rolltheDice` the relevant places. Lines 199-200 can then just be `return rollTheDice()`

script.js
```diff
@@ -1,4 +1,275 @@
...
+      (hasPlayer1Rolled == 1 &&
+        hasPlayer2Rolled &&
+        player2ConcatenatedNumber == 0))
+  ) {
+    return;
+  }
+  if (userInput == "" && !(hasPlayer1Rolled == 1 && hasPlayer2Rolled == 1)) {
+    //Check if there is no input, which means the players are rolling the dice. Call the dice rolling functions
+    rollTheDice();
+    return statementForDiceRolled();
```
lim-jiahao, line 200: see my comment above for the `rollTheDice` function

script.js
```diff
@@ -1,4 +1,275 @@
...
+  }
+
+  //Check if input is "dice 1" or "dice 2", which means the players are choosing which dice is the first number and which is the second. Call the concatenating functions
+  if (userInput == "dice 1" || userInput == "dice 2") {
+    concatenator(userInput);
+    return statementForConcatenation(userInput);
+  }
```
lim-jiahao, line 207: probably not the best from a user experience point of view for the user to have to type 'dice 1' or 'dice 2' to choose the order of the dice. I know the design choice for this is because you eventually do `capitaliser(userInput)` in `statementForConcatenation` to get 'Dice 1' or 'Dice 2' as part of the output msg. 

Personally would have the user just type in 1 or 2 then hard code the word 'Dice' as part of your output msg, then just insert the `userInput` (which would be 1 or 2) to the string. This way you can do without your `capitalizer` function altogether as this is the only place it is used

script.js
```diff
@@ -1,4 +1,275 @@
...
+      "Player 2 Dice 2: ",
+      player2DiceRolls[1]
+    );
+    //Change the global variable hasPlayer1Rolled to 1 to show that Player 1 has rolled
+    hasPlayer2Rolled = 1;
+  }
+};
+
+//Create a function that concatenates the players' two dice numbers
+var concatenator = function (userInput) {
```
lim-jiahao, line 74: This function is ok too but the code for concatenating player 1's rolls and concatenating player 2's rolls is fairly repetitive, consider how you can refactor this. In programming, we try not to repeat ourselves if possible - the reason for this is 1. for readability purpose, and 2. for code maintainability purpose, imagine you notice a bug here and need to change something, then you need to change it twice for the code sections you repeated. And this gets more complex if there are more instances of unnecessary repetition in your code.

script.js
```diff
@@ -1,4 +1,275 @@
...
+
+  //Check if input is "compare", which means its time to compare the two numbers and see who won
+  if (
+    userInput == "compare" &&
+    hasPlayer1Rolled == 1 &&
+    hasPlayer2Rolled == 1
+  ) {
+    timeToReset = 1;
+    return comparingWinnerStatement();
+  }
```
lim-jiahao, line 217: again, probably not the best from a user experience point of view for the user to have to type 'compare' in order to get the outcome of the game. if both players have already rolled their dice and selected their order, then the only next logical progression for the game is to compare the scores. would be better if you just ignore whatever input is given and immediately move on to your `comparingWinnerStatement` function.

script.js
```diff
@@ -1,4 +1,275 @@
...
+  //Check if player 1 won
+  if (player1ConcatenatedNumber > player2ConcatenatedNumber) {
+    return `Player 1's number is ${player1ConcatenatedNumber}. <br>Player 2's number is ${player2ConcatenatedNumber}. <br><br>Player 1 won! <br><br>${howToReset}`;
+  }
+
+  //Check if player 2 won
+  if (player2ConcatenatedNumber > player1ConcatenatedNumber) {
+    return `Player 1's number is ${player1ConcatenatedNumber}. <br>Player 2's number is ${player2ConcatenatedNumber}. <br><br>Player 2 won! <br><br>${howToReset}`;
+  }
+};
```
lim-jiahao, line 169: nice use of early return here to organise the code. because of this use of early return, we do not need the last if in line 166, so this can be just 

```
if (player1ConcatenatedNumber == player2ConcatenatedNumber) {
    return `Both of your numbers are ${player1ConcatenatedNumber}. <br><br>It's a tie! <br><br>${howToReset}`;
  }

//Check if player 1 won
if (player1ConcatenatedNumber > player2ConcatenatedNumber) {
    return `Player 1's number is ${player1ConcatenatedNumber}. <br>Player 2's number is ${player2ConcatenatedNumber}. <br><br>Player 1 won! <br><br>${howToReset}`;
  }

  //player 2 won
return `Player 1's number is ${player1ConcatenatedNumber}. <br>Player 2's number is ${player2ConcatenatedNumber}. <br><br>Player 2 won! <br><br>${howToReset}`;
```
The way the code above works is: if the first condition is met return early, otherwise it carries on to the second condition, again if it is met then we return early. Otherwise we carry on to the last part which does not need to be in an if statement, this serves as our 'default' case if we didn't return early

script.js
```diff
@@ -1,4 +1,275 @@
...
+    hasPlayer1Rolled == 1 &&
+    hasPlayer2Rolled == 1
+  ) {
+    timeToReset = 1;
+    return comparingWinnerStatement();
+  }
+
+  //Check if input is "reset", which means that it is time to reset the game
+  if (userInput == "reset" && timeToReset == 1) {
+    return gameReset();
```
lim-jiahao, line 221: good use of a reset function!

script.js
```diff
@@ -1,4 +1,275 @@
...
+    hasPlayer1Rolled == 1 &&
+    hasPlayer2Rolled == 0 &&
+    player1ConcatenatedNumber == 0
+  ) {
+    return `Player 1, please select which dice you would like to be in front. <br><br>You rolled a ${player1DiceRolls[0]} for Dice 1 and a ${player1DiceRolls[1]} for Dice 2. <br><br> Input "dice 1" for Dice 1 or "dice 2" for Dice 2`;
+  }
+
+  if (hasPlayer1Rolled == 1 && hasPlayer2Rolled == 0) {
+    return `Please press the "submit" button without typing anything in the input box to start the game for player 2`;
+  }
```
lim-jiahao, line 243: same comment as above, do we need this odd constraint of the input box being empty?

lim-jiahao:
Hey Han Chun,

Good job! I got the chance to briefly look at your code during the redux session when you were sharing, and like I mentioned that day, your use of helper functions is amazing and makes the code very neat and organised. It is very well commented as well and this was what you mentioned you are pleased about which I definitely agree with. You showed good grasp of everything taught so far. 

Some high level comments: (some of which I elaborated on in my inline comments)
- Input validation is done in a unique way but might not be optimal, consider incorporating it into the `beatThatGame` function
- Variable names are generally ok but some function names could be better e.g. `concatenator`, what is it concatenating? 
- In some places, code is repetitive so consider refactoring
- Showing some instructions on how to start the game or what this game is about the moment the game loads would be ideal

Keep up the good work for blackjack and I'm excited to see what you come up with!

PR #287, opened 2021-12-11
12-2 Nathan

nathannzq:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

2.5

Comfort Level (1-5):
4
Completeness Level (1-5):
4
Fun and useful


script.js
```diff
@@ -1,5 +1,202 @@
...
+*/
+
+// Global Var
+player1diceResults = []
+player2diceResults = []
+player1 = ""
+player2 = ""
+player1Number = 1
+player2Number = 2
+//
```
lim-jiahao: Nice and clear global variables, do remember to declare them with `var`. There are some side effects of not declaring them with `var` which I won't get into, which is why it is recommended to declare it with `var` even if the code still works the way you have it.

script.js
```diff
@@ -1,5 +1,202 @@
...
+  firstDice = rollDice()
+  secondDice = rollDice()
+  array.push(firstDice)
+  array.push(secondDice)
+  console.log(`roll2Dice: ${firstDice}, ${secondDice}`)
+  console.log(`array: ${array}`)
+  console.log(`player1diceResults: ${player1diceResults}`)
+  console.log(`player2diceResults: ${player2diceResults}`)
+  return array
+}
```
lim-jiahao: Again, as `firstDice` and `secondDice` are not global variables, do define them using `var`

Alternatively you can just do `array = [rollDice(), rollDice()]` and do away with the `firstDice` and `secondDice` variables altogether

script.js
lim-jiahao: However, i see that `player1Number` and `player2Number` are subsequently not used anywhere in your code, seems like a good place to use them is e.g. in lines 160, 167, etc where you pass in 1 or 2 to a function

script.js
```diff
@@ -1,5 +1,202 @@
...
+    myOutputValue = `Issa Draw lubba dubba dub dub <br><br>
+    Click enter to start another new game`;
+    return myOutputValue;
+  }
+};
+
+
+
+
+currentGameMode = "Welcome to Beat That"
```
lim-jiahao: Better if this variable is at the top too with the other global variables

script.js
```diff
@@ -1,5 +1,202 @@
...
+      1. This is a 2 person, player vs player game. <br>
+      2. The game will roll 2 dice and show the dice rolls result. For example, dice 1 rolled 3 and dice 2 rolled 4. <br>
+      3. The player that rolled the dice will pick the order of the dice.(If they wanted the number 43, they would specify that the 2nd dice goes first.) <br>
+      4. After both players have rolled and chosen their dice order, the player with the higher combined number wins.  `;
+
+    currentGameMode = "Enter Player 1 name";
+    console.log(`currentGameMode: ${currentGameMode}`);
+
+    return myOutputValue;
+  }
```
lim-jiahao: this is ok, but try seeing how if you can output the message into the output box immediately on page load by editing the html so that the user sees the message straight away when he loads the game. Currently when I load your game, I am greeted with an empty text input and output box, and it is not immediately clear what I have to do. Always think about user experience when you are designing something!

This will involve editing the html which I realise you didn't have a chance to in this project but do try playing around with HTML/CSS for blackjack! Especially so since you're interested in bootcamp. The first part of bootcamp is mainly front-end doing DOM manipulation using HTML/CSS and JavaScript so it would be good to have some exposure right now, even if it is simple things like changing the background color of the page or the container box, or changing the fonts that are being used.

script.js
```diff
@@ -1,5 +1,202 @@
...
+    return myOutputValue
+  }
+  if (currentGameMode == "Enter Player 2 name"){
+    player2 = input
+    currentGameMode = "Player 1 turn"
+    console.log(`Player 2: ${player1}`);
+    myOutputValue = `Welcome ${player2}. <br> Player 1, press enter to start your roll `;
+    console.log(`currentGameMode: ${currentGameMode}`);
+    return myOutputValue
+  }
```
lim-jiahao: this is ok but a bit repetitive, consider how you can refactor this so that there is less code repetition

one way i can think of is having a `playerNames` array instead of the `player1` and `player2` variables you currently have then keep pushing new player names into the array until the desired amount of players is reached, upon which you switch the game state to 'Player 1 turn'

script.js
```diff
@@ -1,5 +1,202 @@
...
+      Type 2 and click enter if you wish for dice 2 to be the first number`}
+    else if (playerNum == 2){
+      return `Please enter your choice. <br><br> 
+      dice 1 : ${player2diceResults[0]} <br><br>
+      dice 2 : ${player2diceResults[1]} <br><br>
+      Please choose the order of desired dice result. <br>
+      Type 1 and click enter if you wish for dice 1 to be the first number <br>
+      Type 2 and click enter if you wish for dice 2 to be the first number`
+    }
+  }
```
lim-jiahao: again a bit repetitive, consider refactoring

the only difference is whether we are displaying from `player1diceResults` or `player2diceResults`, we prob can create variables called `dice1` and `dice2` to store the relevant value and have one single return

```
   var dice1;
   var dice2;
    if(playerNum == 1){
       dice1 = player1diceResults[0];
       dice2 = player1diceResults[1];
    } else if (playerNum == 2){
       dice1 = player2diceResults[0];
       dice2 = player2diceResults[1];
    }
    return `Please enter your choice. <br><br> 
      dice 1 : $dice1} <br><br>
      dice 2 : ${dice2} <br><br>
      Please choose the order of desired dice result. <br>
      Type 1 and click enter if you wish for dice 1 to be the first number <br>
      Type 2 and click enter if you wish for dice 2 to be the first number`
```

script.js
```diff
@@ -1,5 +1,202 @@
...
+    player2DiceOrder = input
+    myOutputValue = playerTurn(2, player2DiceOrder,player2diceResults)
+    currentGameMode = "Game Outcome"
+    console.log(`currentGameMode: ${currentGameMode}`);
+    return myOutputValue
+  }
+  else if(currentGameMode == "Player 2 choice" && input == ""){
+    currentGameMode = "Player 2 choice"
+    return choiceErrorMsg(2)
+  }
```
lim-jiahao: nice abstraction of the player's turn into a separate helper function, makes this more readable! same point again, a bit repetitive, consider how this can be refactored. notice how lines 159-175 are very similar to lines 176-192, except for the player num. 

the reason why i emphasise the point of repetition so much is because in programming, we try not to repeat ourselves if possible - the reason for this is 1. for readability purpose, and 2. for code maintainability purpose, imagine you notice a bug here and need to change something, then you need to change it twice for the code sections you repeated. And this gets more complex if there are more instances of unnecessary repetition in your code.

script.js
```diff
@@ -1,5 +1,202 @@
...
+  if (currentGameMode == "Game Outcome"){
+    myOutputValue = gameOutcome()
+    console.log(`currentGameMode: ${currentGameMode}`);
+    currentGameMode = "Player 1 turn"
+    // reinitialise array data
+    player1diceResults = []
+    player2diceResults = []
+    return myOutputValue
+  }
```
lim-jiahao: again nice use of a helper function! your main function currently is kinda huge, but my suggestion above on refactoring the code should alleviate it somewhat. for separate if statements such as this one, consider putting line breaks also to make it more readable and clear that its a separate section of the function altogether

script.js
```diff
@@ -1,5 +1,202 @@
...
+  else if (playerNumber == 2 && choice == null){
+    player2diceResults = roll2Dice(player2diceResults)
+    myOutputValue = `Player ${playerNumber}, Here is your rolled dice result: <br><br>
+    dice 1 : ${player2diceResults[0]} <br><br>
+    dice 2 : ${player2diceResults[1]} <br><br>
+    Please choose the order of desired dice result. <br>
+    Type 1 and click enter if you wish for dice 1 to be the first number <br>
+    Type 2 and click enter if you wish for dice 2 to be the first number`;
+    return myOutputValue
+  }
```
lim-jiahao: see the other comments on repetition and how to refactor this

script.js
```diff
@@ -1,5 +1,202 @@
...
+  else if (choice == 2){
+    myOutputValue = `${playerNumber} have chose dice 2 to be the first in order, your result is ${playerArray[1]}${playerArray[0]} <br><br>
+    Click enter for next.`;
+
+    console.log(`player ${playerNumber} dice results: ${playerArray[1]}${playerArray[0]}`);
+    console.log(`user dice order: ${choice}`);
+    return myOutputValue
+  }
```
lim-jiahao: this part prob needs additional input validation. when trying to play your game, if i enter a choice that is not 1 or 2, i get back `undefined` in the output box which is probably not what you want

script.js
```diff
@@ -1,5 +1,202 @@
...
+     ${player2} BEATS ${player1} with a score of ${Number(player2diceResults.join(""))} against ${Number(player1diceResults.join(""))}<br><br>
+    Click enter to start another new game`;
+    return myOutputValue;
+  }
+  else if (Number(player2diceResults.join("")) == Number(player1diceResults.join(""))){
+    myOutputValue = `Issa Draw lubba dubba dub dub <br><br>
+    Click enter to start another new game`;
+    return myOutputValue;
+  }
+};
```
lim-jiahao: this is good but consider storing the `Number(player2diceResults.join("")` and `Number(player1diceResults.join(""))` as a variable before using it in your conditionals and messages. This is for readability purpose and also, if you find a bug in the way you are calculating the score, you only need to change it in 2 places instead of 10 with the current way your code is. 

Also, since you are making use of early return, you do not need to nest the conditions in `else if` and you actually don't even need the last conditional check.

To demonstrate:
```
var gameOutcome = function () {
  var player1Score = Number(player1diceResults.join(""));
  var player2Score = Number(player2diceResults.join(""))
  var myOutputValue = "";
  if (player1Score > player2Score) {
    myOutputValue = `Congratulations ${player1} for winning!!! <br><br>
    ${player1} BEATS ${player2} with a score of ${player1Score} against ${player2Score} <br><br>
    Click enter to start another new game`;
    return myOutputValue;
  } 
  
  if (player2Score > player1Score) {
    myOutputValue = `Congratulations ${player2} for winning!!! <br><br>
     ${player2} BEATS ${player1} with a score of ${player2Score} against ${player1Score}<br><br>
    Click enter to start another new game`;
    return myOutputValue;
  }

  myOutputValue = `Issa Draw lubba dubba dub dub <br><br>
    Click enter to start another new game`;
  return myOutputValue;
};
```

See how the `else if` and the last condition is gone altogether? The way the code above works is: if the first condition is met return early, otherwise it carries on to the second condition (which is why we don't need the else if), again if it is met then we return early. Otherwise we carry on to the last part which does not need to be in a conditional statement, this serves as our 'default' case if we didn't return early.

script.js
lim-jiahao: also your resetting of the game state is fine here since you are only resetting the two dice results array. but consider having a function to reset the game state if you have more state variables to reset, which might be the case for blackjack

lim-jiahao:
Hey Nathan,

Generally very solid code. A lot of my comments really are just about reducing code repetition and suggestions on how to refactor. You demonstrate good understanding of the concepts so far and put it to good use. 

High level comments:
1. Remember to declare your variables with `var` for global variables, or local variables the first time they are used in a function. 
2. Refactor to reduce code repetition and enhance readability + maintainability
3. Try out HTML/CSS to get exposure especially before bootcamp

Hope you get a chance to try out more than just Base for blackjack, as I noticed you only did Base for SPS and Beat That. You are definitely more than capable of doing so and it would be good practice. Excited to see what you come up with!

PR #288, opened 2021-12-11
12-2 Noel

noellimx:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

8

playerWinners is correctly initialized to a list, but it was treated as a number. Ref: 3342c40976390d7c62a5183837d9e5a8ee9ab915

showing the display
Comfort Level (1-5):
2
Completeness Level (1-5):
2.5

The html shown depends on the game state

i'm not sure if rendering is the term, but i name functions which changes the html as renderSmthSmth()

### DIFF
--- a/game-logic.js
+++ b/game-logic.js
@@ -68,7 +68,7 @@ const changePlayer = function () {
 
 /* Dice motions */
 
-const rollDice = function () {
+const rollDie = function () {
   var face = 6;
   return Math.floor(Math.random() * face) + 1;
 };
@@ -216,7 +216,7 @@ const setButtonValue = function (text) {
 /* Main Action */
 
 var actionRoll = function () {
-  playerRoll = [rollDice(), rollDice()]; // roll two dice
+  playerRoll = [rollDie(), rollDie()]; // roll two dice
   playerRolls[CURRENT_PLAYER] = playerRoll; // record playerRoll in playerRolls
   generateOrderings(CURRENT_PLAYER, playerRoll); // record permutations
 
--- a/game-logic.js
+++ b/game-logic.js
@@ -191,7 +191,7 @@ var actionOrder = function (index) {
   if (!isValidRollOrdering(CURRENT_PLAYER, index)) {
     return `Invalid choice "${index}". Please re-enter`;
   }
-
+  index = Number(index);
   playerOrderingChoice[CURRENT_PLAYER] =
     playersRollsOrderings[CURRENT_PLAYER][index]; // record the player choice of ordering
 
--- a/game-logic.js
+++ b/game-logic.js
@@ -73,6 +73,7 @@ const rollDice = function () {
   return Math.floor(Math.random() * face) + 1;
 };
 
+// Permutations
 const generateOrderings = function (player, twoDiceValue) {
   const [val1, val2] = twoDiceValue;
   playersRollsOrderings[player].push([val1, val2]);
@@ -81,6 +82,7 @@ const generateOrderings = function (player, twoDiceValue) {
   }
 };
 
+// How many choices does the player have?
 const getOrderingChoiceCount = function (player) {
   return playersRollsOrderings[player].length;
 };
@@ -170,6 +172,39 @@ const displayCurrentPlayer = function (PLAYER) {
   }
 };
 
+/**
+ *          Player 1    Player 2
+ *  Roll     3            2
+ *  Choice    1           1
+ */
+
+const generateRoundStatisticsTable = function () {
+  var table = document.createElement("table");
+  table.className += ` flex-table`;
+  var headerRow = table.insertRow();
+  headerRow.insertCell().innerHTML = "#(Total Wins W): ";
+  for (let playerIndex = 0; playerIndex < PLAYERS; playerIndex++) {
+    headerRow.insertCell().innerHTML = `${displayCurrentPlayer(
+      playerIndex
+    )}  (${PLAYER_WIN_COUNT[playerIndex]})`;
+  }
+
+  var rollRow = table.insertRow();
+  rollRow.insertCell().innerHTML = "Roll: ";
+  for (let playerIndex = 0; playerIndex < PLAYERS; playerIndex++) {
+    rollRow.insertCell().innerHTML = playerRolls[playerIndex];
+  }
+
+  var choiceOfOrderingRow = table.insertRow();
+  choiceOfOrderingRow.insertCell().innerHTML = "Choice: ";
+  for (let playerIndex = 0; playerIndex < PLAYERS; playerIndex++) {
+    choiceOfOrderingRow.insertCell().innerHTML =
+      playerOrderingChoice[playerIndex];
+  }
+
+  return table;
+};
+
 /* Helpers to set innerHTML */
 const setInputDescription = function (text) {
   HTML_G_INPUT_DESC.innerHTML = text;
@@ -177,6 +212,8 @@ const setInputDescription = function (text) {
 const setButtonValue = function (text) {
   HTML_G_BUTTON.innerHTML = text;
 };
+
+/* Main Action */
 
 var actionRoll = function () {
   playerRoll = [rollDice(), rollDice()]; // roll two dice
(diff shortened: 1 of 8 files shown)


game-logic.js
```diff
@@ -0,0 +1,367 @@
...
+ */
+var playerOrderingChoice = [[], []];
+
+const ACTION_ROLL = 0;
+const ACTION_ORDER = 1;
+const ACTION_GAME_SET = 2;
+
+var CURRENT_ACTION = ACTION_ROLL;
+
+var PREV_ROUND_RESULT_DESCRIPTION = "";
```
lim-jiahao, line 49: nice and clear global variables with good commenting for the more complicated ones (your three 2d arrays)

lim-jiahao: i generally only use SCREAM_CASE for constants and not for global variables that still might change, but i know this is a discussion you had with kenrick before and there is no clear standard for this, so its really up to personal preference

game-logic.js
```diff
@@ -0,0 +1,367 @@
...
+  playerOrderingChoice = [[], []];
+  PREV_ROUND_RESULT_DESCRIPTION = "";
+  return "Game reset";
+};
+const changeAction = function (ACTION) {
+  CURRENT_ACTION = ACTION;
+};
+const changePlayer = function () {
+  CURRENT_PLAYER = (CURRENT_PLAYER + 1) % PLAYERS;
+};
```
lim-jiahao, line 65: these two are nice but i question the value of having these fairly simple one-line assignments as helper functions, seems overly granular to me but i can def see the benefits

game-logic.js
```diff
@@ -0,0 +1,367 @@
...
+
+// Permutations
+const generateOrderings = function (player, twoDiceValue) {
+  const [val1, val2] = twoDiceValue;
+  playersRollsOrderings[player].push([val1, val2]);
+  if (val2 !== val1) {
+    playersRollsOrderings[player].push([val2, val1]);
+  }
+};
```
lim-jiahao, line 83: nice use of destructuring to get out the vals from twoDiceValue

game-logic.js
```diff
@@ -0,0 +1,367 @@
...
+
+// Check if index is in the orderings array of current player
+const isValidRollOrdering = function (PLAYER, index) {
+  if (
+    index === undefined ||
+    index === null ||
+    playersRollsOrderings[index] === undefined
+  ) {
+    return false;
+  }
```
lim-jiahao, line 98: this prob can be reduced to just `playersRollsOrderings[index] === undefined` because if index is null or undefined, then looking up the array index at null or undefined will return undefined anyway, making the condition true

game-logic.js
```diff
@@ -0,0 +1,367 @@
...
+  index = Number(index);
+  if (Number.isNaN(index)) {
+    return false;
+  }
+
+  let maxIndex = playersRollsOrderings[PLAYER].length;
+  return 0 <= index && index < maxIndex;
```
lim-jiahao, line 105: prob can be combined tgt to `return Number.isNaN(index) && 0 <= index && index <= maxIndex`

if its NaN then this condition evaluates to false and stops checking the next two conditions, returning false overall

game-logic.js
```diff
@@ -0,0 +1,367 @@
...
+// result number is joint list of numbers
+// [1,6] => 16
+const concatenatedVal = function (ordering) {
+  var str = ``;
+  for (let i = 0; i < ordering.length; i += 1) {
+    str += `${ordering[i]}`;
+  }
+
+  var concatVal = Number(str);
+  return Number(concatVal);
```
lim-jiahao, line 117: prob only need to cast it to Number once?

game-logic.js
```diff
@@ -0,0 +1,367 @@
...
+};
+
+// This function should be called when all player had chosen their ordering.
+const decideRoundWinner = function () {
+  var max = 0;
+  var playerWinners = [];
+  for (
+    let playerIndex = 0;
+    playerIndex < playerOrderingChoice.length;
+    playerIndex++
```
lim-jiahao, line 127: convention is to use += 1 instead of ++ but no problem if you wanna stick to ++, just be consistent as i noticed you used += 1 above

if you're curious why: 
https://eslint.org/docs/rules/no-plusplus

game-logic.js
```diff
@@ -0,0 +1,367 @@
...
+  return table;
+};
+
+/* Helpers to set innerHTML */
+const setInputDescription = function (text) {
+  HTML_G_INPUT_DESC.innerHTML = text;
+};
+const setButtonValue = function (text) {
+  HTML_G_BUTTON.innerHTML = text;
+};
```
lim-jiahao, line 220: again, same comment as above for helper functions with simple one-line assignments. in this case i understand it more because you prob wish to abstract away the DOM manipulation from the core function, but i think the fact that you already have your DOM element stored in variables makes it looks clean enough as is

game-logic.js
```diff
@@ -0,0 +1,367 @@
...
+    result += `Previous activity: ${actionOrder(value)}`;
+  } else if (CURRENT_ACTION === ACTION_GAME_SET) {
+    result += actionGameSet();
+  }
+  nextRenderParagraphMain(result);
+  nextRenderParagraphRound();
+};
+
+// nextRenderParagraphMain should be called AFTER the next action state has been updated.
+var nextRenderParagraphMain = function (result) {
```
lim-jiahao, line 307: your DOM manipulation in this function is solid, but personal opinion its not the most readable. i understand your intentions to abstract away all the DOM manipulation by having the helper functions and the DOM elements stored in variables that are declared in another file. but when reading this I gotta constantly reference the other file to see exactly what element this is and then to the function definition to see exactly what it does, while taking into consideration the DOM elements it takes in

mainParagraph.js
```diff
@@ -0,0 +1,33 @@
...
+
+const ID_HTML_SUBMIT_BUTTION = "submit-button";
+const ID_HTML_INPUT_DESCRIPTION = "input-description";
+const ID_HTML_INPUT_FIELD = "input-field";
+const ID_HTML_OUTPUT_DESCRIPTION = "output-description";
+const ID_HTML_PARAGRAPH_MAIN = "paragraph-main";
+const CLASS_NAME_HTML_FLEX_DIV = "flex-div";
+
+const HTML_G_PARAGRAPH_MAIN = document.createElement("paragraph");
+HTML_G_PARAGRAPH_MAIN.setAttribute("id", ID_HTML_PARAGRAPH_MAIN);
```
lim-jiahao, line 15: there isn't a `paragraph` element in HTML, did you mean `p`? regardless, you are adding children to this element subsequently in your code, such as `div`s. for semantic HTML, a `p` element is probably not the best choice here, and a `div` would probably be better

lim-jiahao:
Hey Noel,

Great work as usual and I really don't have much comments lol, that's why most of them are on the nitpicky side. Your gameplay logic is solid and I have not much comments there. Very good understanding of all the fundamental concepts which you put to good use.

The only comment I would really emphasise is the comment on `nextRenderParagraphMain`. Again, I see the benefits in doing the way you did but see if you can refactor it to make it less complex. Nevertheless, very good job in trying out HTML/CSS before bootcamp. And if it is one of your first few attempts at doing so, then I think you did a very good job already. Like I said before, the first part of bootcamp is mainly front-end doing DOM manipulation using HTML/CSS and JavaScript so it would be good to have some exposure right now, even if it is simple things like changing the background color of the page or the container box, or changing the fonts that are being used. So you are definitely positioning yourself well even before bootcamp starts. Keep it up for blackjack! 

Would be nice if you can try some of the More Comfortable as I think it would be a good way to challenge yourself further. If you have time for blackjack, definitely do so.

game-logic.js
noellimx: i may have confused globals with constants  :sweat_smile:

mainParagraph.js
noellimx: noted

noellimx:
thanks for your review and pointing out the inconsistencies :) 

for `nextRenderParagraphMain`, definitely the variables can be nearer to where it's used! :see_no_evil: am working on how to organize the files in a "tidy" way ...

PR #291, opened 2021-12-11
Grace:BeatThat

gracethedragon:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

What part of the assignment did you spend the most time on? **Figuring how how to not accept inputs other than what I want**

Comfort Level (1-5): **3**

Completeness Level (1-5): **3**

What did you think of this deliverable? **Interesting**

What's one aspect of your code you would like specific, elaborate feedback on? **How to ensure only specific inputs are allowed**

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,11 +1,22 @@
+//1. Player 1 to roll dice
+//2. Player 1 to decide order of dice
+//3. Player 2 to roll dice
+//4. Player 2 to decide order of dice
+//5. Determine if Player 1 or 2 wins
+//6. Game continues with running score being tabulated
+
 var diceRoll_1 = "";
 var diceRoll_2 = "";
 var diceResult = "";
-var player = "";
+var player = "Player 1";
 var winner = "";
+var gameMode = "start";
+var myOutputValue = "";
 var player1score = [];
 var player2score = [];
-var gameMode = "play";
+var player1runningscore = "";
+var player2runningscore = "";
+var leaderboard = "";
 
 var diceRoll = function (input) {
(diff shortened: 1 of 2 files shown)


script.js
```diff
@@ -1,4 +1,113 @@
...
+var diceRoll = function (input) {
+  if (input == "roll") {
+    diceRoll_1 = Math.floor(Math.random() * 6) + 1;
+    diceRoll_2 = Math.floor(Math.random() * 6) + 1;
+  }
+};
+
+var diceCombined = function (input) {
+  if (input == 1) {
+    diceResult = parseInt(diceRoll_1.toString() + diceRoll_2.toString());
```
daves77, line 30: diceRoll1 & diceRoll_2 should already be an integer, so type conversion is not necessary here!

script.js
```diff
@@ -1,4 +1,113 @@
...
+      } else if (player == "Player 2") {
+        player2score.push(diceResult);
+        gameMode = "score";
+        myOutputValue = `Press "submit" for results`;
+      }
+    }
+    return myOutputValue;
+  }
+
+  while (gameMode == "score") {
```
daves77, line 83: you don't have to use while here! you can use "if" instead!

daves77: while is unnecessary here as you're changing gameMode @ line 110, so an "if" statement is sufficient!

daves77 (commented):
your code is clean concise, keep it up! do note the differences between while's and if's to check for conditionals though!

PR #292, opened 2021-12-11
Beat That Base Submission.

SK-Luke:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

Ans: 2.5 hours

Ans: My dice roll results kept changing everything I click submit. Solved it by adding results of role into array so that the dice will not roll again to change result.

Ans: Refactoring the helper function

Comfort Level (1-5):
4

Completeness Level (1-5):
5

Ans: It is doable and interesting as we are able to incorporate everything we have learnt so far.

Ans: I managed to finish the project, and had use consolo.log more often to ensure every step is well.

Ans: 
1. Naming of my variables. I feel that some naming might be too long, or seems conflicting with other names as it looks similar.

2. I did not reset player names at the end of the game. Is it better to reset? Since player names will be assigned a new value every time the game restarts.

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,294 @@
+/* Beat That - Base
+Requirements
+1. There are 2 players and players take turns.
+2. When a player clicks Submit, the game rolls 2 dice and shows the dice rolls, for example 3 and 6.
+3. The player picks the order of the dice they want. For example, if they wanted the number 63, they would specify that the 2nd dice goes first. You can choose how the player        
+   specifies dice order.
+4. After both players have rolled and chosen dice order, the player with the higher combined number wins.
+
+Objectives:
+1. 1v1 player vs player dice roll game &&
+2. Roll 2 dice and concentate the 2 dice number together &&
+3. Players can choose order of dice to concentate (dice 1 + dice 2 || dice 2 + dice 1)
+
+What is required?
+1. Dice Roll helper function
+2. Global states for different gamemode
+3. Game function
+4. Instructions for user
+*/
+
+//GLOBAL VARIABLES
+var currentGameMode = "welcomeToBeatThat"; //Starting Game Mode
+var player1 = ""; //Player 1 IGN
+var player2 = ""; //Player 2 IGN
+
+var userChosenDiceOrder = ""; //User desired dice order, either 1 or 2 (input = userDiceOrder)
+
+// player 1 dice results, store in array for later use, prevent another diceroll to reset results too.
+var player1dice1Arr = [];
+var player1dice2Arr = [];
+// player 2 dice results, store in array for later use, prevent another diceroll to reset results too.
+var player2dice1Arr = [];
+var player2dice2Arr = [];
+
+var player1Score = ""; //Player 1 score at the end of turn
+var player2Score = ""; //Player 2 score at the end of turn
+
+// Dice Roll helper function
+var rollDice = function () {
+  // produces a float between 0 and 6
+  var randomFloat = Math.random() * 6;
+  // take off the decimal
+  var resultInteger = Math.floor(randomFloat) + 1;
+  console.log(`rollDice function, ${resultInteger}`);
+  return resultInteger;
+};
+
+// Player 1 Result Function
+var player1Result = function (player1, player2, userChosenDiceOrder) {
+  var myOutputValue = "";
+  if (currentGameMode == "player1Roll") {
+    var diceRoll1 = rollDice();
+    player1dice1Arr.push(diceRoll1);
+    console.log(`dice roll 1 : ${diceRoll1}`);
+    console.log(`player 1 dice 1 array, ${player1dice1Arr}`);
+
+    var diceRoll2 = rollDice();
+    player1dice2Arr.push(diceRoll2);
+    console.log(`dice roll 2 : ${diceRoll2}`);
+    console.log(`player 1 dice 2 array, ${player1dice2Arr}`);
+
+    //Summary Result Message + Note on how to choose dice order
+    myOutputValue = `Your rolled dice result: <br><br>
+  dice 1 : ${player1dice1Arr} <br><br>
+  dice 2 : ${player1dice2Arr} <br><br>
+  Please choose the order of desired dice result. <br>
+  Type 1 and click enter if you wish for dice 1 to be the first number <br>
+  Type 2 and click enter if you wish for dice 2 to be the first number`;
+
+    currentGameMode = "player1Choice"; //Change game mode to allow player 1 to decide dice order
+    console.log(`diceResult of both dice rolled, ${myOutputValue}`);
+  }
+  // Chosen dice 1 to be first
+  else if (currentGameMode == "player1Choice" && userChosenDiceOrder == 1) {
+    myOutputValue = `${player1} have chose dice 1 to be the first in order, your result is ${player1dice1Arr}${player1dice2Arr} <br><br>
+     ${player2}, you may now begin to roll the dice by clicking enter.`;
+
+    player1Score = `${player1dice1Arr}${player1dice2Arr}`; // Player 1 score
+
+    console.log(`player 1 score, ${player1Score}`);
+    console.log(`user Chosen Dice Order, ${userChosenDiceOrder}`);
+    console.log(`Dice 1 as first number results, ${myOutputValue}`);
+
+    currentGameMode = "player2Roll"; //Change game mode to player 2 turn to roll
+    console.log(
+      `Player 1 turn end, now is player 2 turn. Current mode : ${currentGameMode}`
+    );
+
+    return myOutputValue;
+  } else if (currentGameMode == "player1Choice" && userChosenDiceOrder == 2) {
+    myOutputValue = `${player1} have chose dice 2 to be the first in order, your result is ${player1dice2Arr}${player1dice1Arr} <br><br>
+     ${player2}, you may now begin to roll the dice by clicking enter.`;
+
+    player1Score = `${player1dice2Arr}${player1dice1Arr}`; // Player 1 score
+
+    console.log(`player 1 score, ${player1Score}`);
+    console.log(`user Chosen Dice Order, ${userChosenDiceOrder}`);
+    console.log(`Dice 2 as first number results, ${myOutputValue}`);
+
+    currentGameMode = "player2Roll"; // Change game mode to player 2 turn to roll
+    console.log(
+      `Player 1 turn end, now is player 2 turn. Current mode : ${currentGameMode}`
+    );
+  } else if (
+    currentGameMode == "player1Choice" &&
+    (userChosenDiceOrder !== 2 || userChosenDiceOrder !== 1)
+  ) {
+    myOutputValue = `You have enter an invalid input for dice order. Pleaser enter either a 1 or 2 to choose your dice order. Thank you! <br><br>
+    Your rolled dice result: <br><br>
+  dice 1 : ${player1dice1Arr} <br><br>
+  dice 2 : ${player1dice2Arr} <br><br>
+  Please choose the order of desired dice result. <br>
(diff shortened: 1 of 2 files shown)


script.js
```diff
@@ -1,4 +1,294 @@
...
+
+// player 1 dice results, store in array for later use, prevent another diceroll to reset results too.
+var player1dice1Arr = [];
+var player1dice2Arr = [];
+// player 2 dice results, store in array for later use, prevent another diceroll to reset results too.
+var player2dice1Arr = [];
+var player2dice2Arr = [];
+
+var player1Score = ""; //Player 1 score at the end of turn
+var player2Score = ""; //Player 2 score at the end of turn
```
lim-jiahao, line 36: nice and clear global variables, well commented. i would say that if the variable name is descriptive enough, such as the case of `player1Score`, the comment is unnecessary. this is another point for why descriptive variable names are desired as your code becomes self-documenting

i don't see the point of having a `player1dice1Arr` and `player1dice2Arr`, similarly for player 2. seems like in the rest of your code, these arrays only store a max of 1 value, which defeats the purpose of an array. maybe it was an intentional design choice but from the rest of your code, it seems that it would be better if we have two arrays, `player1Rolls` and `player2Rolls`, and we push the dice rolls into the relevant array for each player

script.js
```diff
@@ -1,4 +1,294 @@
...
+      Game Rules: <br>
+      1. This is a 2 person, player vs player game. <br>
+2. The game will roll 2 dice and show the dice rolls result. For example, dice 1 rolled 3 and dice 2 rolled 4. <br>
+3. The player that rolled the dice will pick the order of the dice.(If they wanted the number 43, they would specify that the 2nd dice goes first.) <br>
+4. After both players have rolled and chosen their dice order, the player with the higher combined number wins.  `;
+
+    currentGameMode = "enterPlayer1";
+
+    return myOutputValue;
+  }
```
lim-jiahao, line 236: this is ok, but try seeing if you can output the message into the output box immediately on page load by editing the html so that the user sees the message straight away when he loads the game. Currently when I load your game, I am greeted with an empty text input and output box, and it is not immediately clear what I have to do. Always think about user experience when you are designing something!

This will involve editing the html which I realise you attempted a bit in this project by changing the header and button text but do try playing around with HTML/CSS more for blackjack, even if it is simple things like changing the background color of the page or the container box, or changing the fonts that are being used.

script.js
```diff
@@ -1,4 +1,294 @@
...
+
+    //Reset both player dice array
+    player1dice1Arr = [];
+    player1dice2Arr = [];
+    player2dice1Arr = [];
+    player2dice2Arr = [];
+
+    //Reset player scores
+    player1Score = ""; //Player 1 score at the end of turn
+    player2Score = ""; //Player 2 score at the end of turn
```
lim-jiahao, line 289: a separate helper function to reset your game would prob be ideal, esp if its a more complex game with more variables

for your desired specific feedback, prob not necessary to reset the player names if you are absolutely certain in your gameplay logic that it will get reset to new values when the game restarts. but otherwise, the guideline is better safe than sorry

script.js
```diff
@@ -1,4 +1,294 @@
...
+    //Set player 2 IGN
+    player2 = input;
+
+    currentGameMode = "player1Roll"; //Switch to Game Result for player 1
+
+    console.log(`currentGameMode, ${currentGameMode}`);
+    console.log(`Player 2 IGN, player 2: ${player2}`);
+
+    myOutputValue = `Welcome ${player2}. <br><br>
+    ${player1}, you may begin to start the game by clicking enter.`;
```
lim-jiahao, line 257: this is ok but a bit repetitive, consider how you can refactor this so that there is less code repetition

one way i can think of is having a `playerNames` array instead of the `player1` and `player2` variables you currently have, then just have one game state to keep pushing new player names into the array until the desired amount of players is reached, upon which you switch the game state to 'player1Roll`

script.js
```diff
@@ -1,4 +1,294 @@
...
+    myOutputValue = player1Result(player1, player2, userChosenDiceOrder);
+    return myOutputValue;
+  } else if (
+    //player 2 turn + player 2 result
+    currentGameMode == "player2Roll" || //Dice results of player 2
+    currentGameMode == "player2Choice" //player 2 desired outcome
+  ) {
+    userChosenDiceOrder = input;
+    myOutputValue = player2Result(player2, userChosenDiceOrder);
+    return myOutputValue;
```
lim-jiahao, line 274: similar to above, a bit repetitive, consider how to refactor this into just a single `else if` block

script.js
```diff
@@ -1,4 +1,294 @@
...
+  // produces a float between 0 and 6
+  var randomFloat = Math.random() * 6;
+  // take off the decimal
+  var resultInteger = Math.floor(randomFloat) + 1;
+  console.log(`rollDice function, ${resultInteger}`);
+  return resultInteger;
+};
+
+// Player 1 Result Function
+var player1Result = function (player1, player2, userChosenDiceOrder) {
```
lim-jiahao, line 49: the logic in your `player1Results` and `player2Results` function is very repetitive. think of how you can combine them into one function where in the function, you can detect whose turn it is and by extension, which array you should manipulate or get values from. this will shorten and organise your code greatly

script.js
```diff
@@ -1,4 +1,294 @@
...
+  if (currentGameMode == "player1Roll") {
+    var diceRoll1 = rollDice();
+    player1dice1Arr.push(diceRoll1);
+    console.log(`dice roll 1 : ${diceRoll1}`);
+    console.log(`player 1 dice 1 array, ${player1dice1Arr}`);
+
+    var diceRoll2 = rollDice();
+    player1dice2Arr.push(diceRoll2);
+    console.log(`dice roll 2 : ${diceRoll2}`);
+    console.log(`player 1 dice 2 array, ${player1dice2Arr}`);
```
lim-jiahao, line 60: as per my suggestion above to just have an overall `player1Rolls` array, you can do `player1Rolls = [rollDice(), rollDice()]` here

script.js
```diff
@@ -1,4 +1,294 @@
...
+
+    //Summary Result Message + Note on how to choose dice order
+    myOutputValue = `Your rolled dice result: <br><br>
+  dice 1 : ${player1dice1Arr} <br><br>
+  dice 2 : ${player1dice2Arr} <br><br>
```
lim-jiahao, line 65: this would then be `player1Rolls[0]` and `player1Rolls[1]` respectively

script.js
lim-jiahao: the reason why i emphasise the point of repetition so much is because in programming, we try not to repeat ourselves if possible - the reason for this is 1. for readability purpose, and 2. for code maintainability purpose, imagine you notice a bug here and need to change something, then you need to change it twice for the code sections you repeated. And this gets more complex if there are more instances of unnecessary repetition in your code.

script.js
```diff
@@ -1,4 +1,294 @@
...
+    player1Score = `${player1dice2Arr}${player1dice1Arr}`; // Player 1 score
+
+    console.log(`player 1 score, ${player1Score}`);
+    console.log(`user Chosen Dice Order, ${userChosenDiceOrder}`);
+    console.log(`Dice 2 as first number results, ${myOutputValue}`);
+
+    currentGameMode = "player2Roll"; // Change game mode to player 2 turn to roll
+    console.log(
+      `Player 1 turn end, now is player 2 turn. Current mode : ${currentGameMode}`
+    );
```
lim-jiahao, line 103: even within `player1Result`, this is fairly repetitive as well. notice how the only difference is really line 78 and line 94. one way i would improve this is to compute the score depending on if the dice order is 1 or 2, then construct your output and switch game mode.

```
  else if (currentGameMode == "player1Choice") {
    if (userChosenDiceOrder == 1) {
      player1Score = `${player1dice1Arr}${player1dice2Arr}`;
    } else {
      player1Score = `${player1dice2Arr}${player1dice1Arr}`;
    }

    myOutputValue = `${player1} have chose dice 1 to be the first in order, your result is ${player1Score} <br><br>
     ${player2}, you may now begin to roll the dice by clicking enter.`;

    currentGameMode = "player2Roll"; //Change game mode to player 2 turn to roll
    return myOutputValue;
  }
```

Removed the console.log statements as well to reduce clutter, which is generally good practice too. console.logs are good when you're working on code but when we deploy to production we usually omit them

script.js
```diff
@@ -1,4 +1,294 @@
...
+  dice 2 : ${player1dice2Arr} <br><br>
+  Please choose the order of desired dice result. <br>
+  Type 1 and click enter if you wish for dice 1 to be the first number <br>
+  Type 2 and click enter if you wish for dice 2 to be the first number
+    `;
+
+    console.log(
+      `Invalid input, current mode should still be player choice : ${currentGameMode}`
+    );
+    return myOutputValue;
```
lim-jiahao, line 120: this is fine but i would probably put the negative case up top first and check for that, returning early if so before moving on with the rest of my code

i like that you display the invalid input message to the user along with the dice rolls again! a comment i have for many students is that when they output this msg, they don't show the rolls again and it is very possible the user forgot the dice rolls

script.js
lim-jiahao: also consider having this part in its own function

Your gameplay logic and code can be broken into three distinct parts: 1. players roll, 2. players select order, 3. compare score. This provides a very nice starting point on how we can split the game into separate functions.

you already have the 3rd one with your `gameOutcome` function, but your players roll and players select order is currently together in one function, and you have one function for every player. this is not scalable esp if we introduce more players. Think of how you can come up with say, a `playerRoll` and `playerSelectOrder` function that works for every player

script.js
```diff
@@ -1,4 +1,294 @@
...
+  }
+  return myOutputValue;
+};
+// Player 2 Result Function
+var player2Result = function (player2, userChosenDiceOrder) {
```
lim-jiahao, line 125: see my comments on `player1Result` above to see how this can be improved and refactored

script.js
```diff
@@ -1,4 +1,294 @@
...
+
+    myOutputValue = `Welcome ${player2}. <br><br>
+    ${player1}, you may begin to start the game by clicking enter.`;
+  } else if (
+    // Condition for player 1 turn
+    //player 1 turn + player 1 result
+    currentGameMode == "player1Roll" || //Dice results of player 1
+    currentGameMode == "player1Choice" //player 1 desired outcome
+  ) {
+    userChosenDiceOrder = input;
```
lim-jiahao, line 264: if the dice rolls are the same, does it make sense from a user experience pov to have the user choose the order? think about how we can skip this step and automatically compute the dice roll if they are the same

script.js
```diff
@@ -1,4 +1,294 @@
...
+    ${player1} BEATS ${player2} with a score of ${player1Score} against ${player2Score} <br><br>
+    Click enter to start another new game`;
+    return myOutputValue;
+  } else if (player2Score > player1Score) {
+    myOutputValue = `Congratz ${player2} for winning!!! <br><br>
+     ${player2} BEATS ${player1} with a score of ${player2Score} against ${player1Score}<br><br>
+    Click enter to start another new game`;
+    return myOutputValue;
+  }
+};
```
lim-jiahao, line 215: hmm what if its a tie

lim-jiahao:
Hey Sek Kun,

Good work! A lot of my comments really are just about reducing code repetition and suggestions on how to refactor. You demonstrate good understanding of the concepts so far and put it to good use. The only obvious point of improvement would be to have less repetition in your code.

High level comments:
- Nice attempt at some basic HTML editing, hope to see more of this + CSS in blackjack!
- Refactor to reduce code repetition and enhance readability + maintainability
- For your desired specific feedback, the variable names look ok to me. Which do you feel are very similar?

Hope you get a chance to try out more than just Base for blackjack. You are definitely more than capable of doing so and it would be good practice. Excited to see what you come up with!

PR #294, opened 2021-12-11
B12S01 HONG-FOO-LIANG-Beat-that

fooliang:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

How many hours did you spend on this assignment? 1.5hr

Nothing specific, trial and error around logic required.

Comfort Level (1-5): 4.5

Completeness Level (1-5): 4

What did you think of this deliverable? Manageable logical task

Is there anything in this code that you feel pleased about? Solving problem

html maybe can learn more

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,111 @@
+var Dice1;
+var Dice2;
+var orderMode="";
+var playerMode=1;
+var player1Score=0;
+var player2Score=0;
+var outputmsg1= "Player 1, you chose Dice 1 first.<br>" + "Your number is ";
+var outputmsg2= "Player 1, you chose Dice 2 first.<br>" + "Your number is ";
+var outputmsg3= "Player 2, you chose Dice 1 first.<br>" + "Your number is ";
+var outputmsg4= "Player 2, you chose Dice 2 first.<br>" + "Your number is ";
+var outputmsg5= "<br> It is now Player 1's turn. <br>";
+var outputmsg6= "<br> It is now Player 2's turn. <br>";
+var outputmsg7="The current leaderboard is Player 1 wins with score ";
+var outputmsg8="The current leaderboard is Player 2 wins with score " ;
+var outputmsg9="The current leaderboard is both players tie with score ";
+var outputmsg10=".<br> The Player 1's score is ";
+var outputmsg11=".<br> The Player 2's score is "
+
 var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
+  
+  if (orderMode=="")
+  { Dice1=rollDice();
+    Dice2=rollDice();
+    orderMode="Choose Mode";
+    return "Welcome Player "+ playerMode +". You rolled " + Dice1 +" for Dice 1 and " + Dice2 + " for Dice 2. <br> Choose the order of the dice, eg input 1 for Dice 1, input 2 for Dice 2.";
+  }
+  else if (orderMode =="Choose Mode")
+  {   var combinedNum;
+    if (input==1 && playerMode==1)
+      { combinedNum=Dice1.toString()+Dice2.toString();
+        combinedNum=Math.floor(combinedNum);
+        player1Score=player1Score+combinedNum;
+        orderMode="";
+        playerMode=2;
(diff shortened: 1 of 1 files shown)


script.js
```diff
@@ -1,4 +1,111 @@
+var Dice1;
+var Dice2;
+var orderMode="";
+var playerMode=1;
+var player1Score=0;
+var player2Score=0;
+var outputmsg1= "Player 1, you chose Dice 1 first.<br>" + "Your number is ";
```
jeremylim91, line 7: Notice how there's a lot of repeated text in these lines? (lines 7-9). This is an indication that we can actually use variables to vary certain parts of this string, so that we just need to write the string once rather than repeating it multiple times. One way to do this would be to create a function that takes in several parameters. The function would then return a string with certain parts of the string that are the parameters. Does this make sense? Do reach out if unclear! :)

script.js
```diff
@@ -1,4 +1,111 @@
...
+        combinedNum=Math.floor(combinedNum);
+        player2Score=player2Score+combinedNum;
+        orderMode="";
+        playerMode=1;
+         if (player1Score>player2Score)
+        {  return outputmsg3 + combinedNum + outputmsg5
+        + outputmsg7 +player1Score +outputmsg11 + player2Score +"."; 
+        }
+        else if (player2Score>player1Score)
+        {  return outputmsg3 + combinedNum + outputmsg5
```
jeremylim91, line 77: Try thinking of more semantic variable names that itself describe what the contents are. In this case, in 3 weeks when you read your code you might not remember what output3 contains, and you would have to scroll up to see its contents.

script.js
```diff
@@ -1,4 +1,111 @@
...
+  else if (orderMode =="Choose Mode")
+  {   var combinedNum;
+    if (input==1 && playerMode==1)
+      { combinedNum=Dice1.toString()+Dice2.toString();
+        combinedNum=Math.floor(combinedNum);
+        player1Score=player1Score+combinedNum;
+        orderMode="";
+        playerMode=2;
+        if (player1Score>player2Score)
+        {  return outputmsg1 + combinedNum + outputmsg6
```
jeremylim91, line 36: Can consider using template literals `${variableName}` instead of concatenating the string. Not that any is better or worse, but template literals are just neater to read (if you prefer them)

script.js
```diff
@@ -1,4 +1,111 @@
...
+        + outputmsg7 +player1Score + outputmsg11 + player2Score +"."; 
+        }
+        else if (player2Score>player1Score)
+        {  return outputmsg1 + combinedNum + outputmsg6
+        + outputmsg8 +player2Score + outputmsg10 + player1Score +".";}
+        else
```
jeremylim91, line 42: Good use for else as a catch-all

script.js
```diff
@@ -1,4 +1,111 @@
...
+var outputmsg4= "Player 2, you chose Dice 2 first.<br>" + "Your number is ";
+var outputmsg5= "<br> It is now Player 1's turn. <br>";
+var outputmsg6= "<br> It is now Player 2's turn. <br>";
+var outputmsg7="The current leaderboard is Player 1 wins with score ";
+var outputmsg8="The current leaderboard is Player 2 wins with score " ;
+var outputmsg9="The current leaderboard is both players tie with score ";
+var outputmsg10=".<br> The Player 1's score is ";
+var outputmsg11=".<br> The Player 2's score is "
+
 var main = function (input) {
```
jeremylim91, line 19: Would be good if you can add more comments to explain your line of thought

jeremylim91 (commented):
Heyy @fooliang, good job completing Beat that! I know it's not easy, but the hard work has paid off. As for code quality, one thing you can do is to refactor some of the code so that you can abstract the logic from main into a helper function.

Keep up the good work,look forward to see your Blackjack project~

script.js
fooliang: Take turn for each player to roll two dice and then allow the player to decide which dice comes first to form a two-digit number. Accumulate every time each player throws a dice and compare the the results and show leaderboard.

PR #296, opened 2021-12-11
12:2 Olivia

itsninenow:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

4hrs

undefined. unidentified bracket in x line

debugging what was undefined and where the brackets are 
Comfort Level (1-5):
2
Completeness Level (1-5):
2
Had to do a lot of referencing. Not confident in doing without references.

I tried to take into consideration the user experience and gave feedback in the output messages.

How to select number of players (e.g. 2 players, 3 players, 4 players)

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,150 @@
+/*Requirements
+There are 2 players and players take turns.
+When a player clicks Submit, the game rolls 2 dice and shows the dice rolls, for example 3 and 6.
+The player picks the order of the dice they want. For example, if they wanted the number 63, they would specify that the 2nd dice goes first. You can choose how the player specifies dice order.
+After both players have rolled and chosen dice order, the player with the higher combined number wins.*/
+
+// Objective: Decide who wins the dice rolls by having the higher combined number
+//How to get there? Input roll dice function
+
+// Game modes
+var GAME_MODE_DICE_ROLL = "GAME_MODE_DICE_ROLL";
+var GAME_MODE_CHOOSE_DICE_ORDER = "GAME_MODE_CHOOSE_DICE_ORDER";
+
+// Initialise the game to start with the dice roll game mode
+var gameMode = GAME_MODE_DICE_ROLL;
+
+// Keep track of the current player's number, either 1 or 2.
+// The game starts with Player 1.
+var currPlayer = 1;
+
+// Keep track of each player's dice rolls
+var player1Dice = [];
+var player2Dice = [];
+
+// Keep track of each player's chosen numbers
+var player1Num;
+var player2Num;
+
+// roll dice function
+var getDiceRoll = function () {
+  // Generate a decimal from 0 through 6, inclusive of 0 and exclusive of 6.
+  var randomDecimal = Math.random() * 6;
+
+  // Remove the decimal with the floor operation.
+  // This will be an integer from 0 to 5 inclusive.
+  var randomInteger = Math.floor(randomDecimal);
+
+  // Add 1 to get valid dice rolls of 1 through 6 inclusive.
+  var diceNumber = randomInteger + 1;
+
+  return diceNumber;
+};
+
+/**
+ * Get dice rolls for curr player and populate curr player's dice array
+ * Return the new dice rolls
+ */
+
+var getDiceRolls = function () {
+  // Create an array newDiceRolls with 2 independent dice roll values
+  var newDiceRolls = [getDiceRoll(), getDiceRoll()];
+
+  // Assign newDiceRolls to the current player's dice array
+  if (currPlayer == 1) {
+    player1Dice = newDiceRolls;
+  }
+  // If currPlayer is not 1, assume currPlayer is 2
+  else {
+    player2Dice = newDiceRolls;
+  }
+
+  // Return new dice rolls to parent function
+  return newDiceRolls;
+};
+
+var getPlayerNumber = function (firstNumeralIndex) {
+  // Get the current player's dice array. i.e. var diceArray
+  // if (currPlayer == 1) { diceArray = player1Dice; } else { diceArray = player2Dice; }
+  var diceArray = currPlayer == 1 ? player1Dice : player2Dice;
+  var playerNum;
+  // If the chosen first numeral index is 1, create player number starting with 1st dice
+  if (firstNumeralIndex == 1) {
+    playerNum = Number(String(diceArray[0]) + String(diceArray[1]));
+  }
+  // Otherwise, create player number starting with 2nd dice
+  else {
+    playerNum = Number(String(diceArray[1]) + String(diceArray[0]));
+  }
+
+  // Store player num in the relevant global player num variable
+  if (currPlayer == 1) {
+    player1Num = playerNum;
+  } else {
+    player2Num = playerNum;
+  }
+
+  // Return generated player num to parent function
+  return playerNum;
+};
+
+/**
+ * Compute the winner between Player 1 and Player 2.
+ * Return either 1 or 2 to represent the winning player.
+ * In the event of a tie, Player 2 wins.
+ */
+var determineWinner = function () {
+  if (player1Num > player2Num) {
+    return 1;
+  }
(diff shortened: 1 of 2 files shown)


script.js
```diff
@@ -1,4 +1,150 @@
...
+// The game starts with Player 1.
+var currPlayer = 1;
+
+// Keep track of each player's dice rolls
+var player1Dice = [];
+var player2Dice = [];
+
+// Keep track of each player's chosen numbers
+var player1Num;
+var player2Num;
```
lim-jiahao, line 27: nice and clear global variables, well commented. i would say that if the variable name is descriptive enough, such as the case of `player1Dice`, the comment is unnecessary. this is another point for why descriptive variable names are desired as your code becomes self-documenting

i like that you're using scream case for the first two variable to denote that they won't change. Consider using the `const` keyword to declare them too rather than `var`. For example `const GAME_MODE_DICE_ROLL = "GAME_MODE_DICE_ROLL"`. I know this wasn't really covered in the course but it's a minor step up if you're interested.

For more info:

script.js
```diff
@@ -1,4 +1,150 @@
...
+    player1Dice = newDiceRolls;
+  }
+  // If currPlayer is not 1, assume currPlayer is 2
+  else {
+    player2Dice = newDiceRolls;
+  }
+
+  // Return new dice rolls to parent function
+  return newDiceRolls;
+};
```
lim-jiahao, line 64: this is beautiful and such a smart way to assign dice rolls to the global arrays depending on whose turn it is

i suggested similar ways to the other students who have more repetitive code to push dice rolls into the player1 and player2 array so its nice to see it being implemented here, good work!

since you are updating the global variable, you probably don't need to return so line 63 can be omitted. when this is called in line 107, you can just do `getDiceRolls()` without having to assign it to anything. functions can return a value but this is by no means a necessity and in your case its an example of it not being necessary

script.js
```diff
@@ -1,4 +1,150 @@
...
+
+var getPlayerNumber = function (firstNumeralIndex) {
+  // Get the current player's dice array. i.e. var diceArray
+  // if (currPlayer == 1) { diceArray = player1Dice; } else { diceArray = player2Dice; }
+  var diceArray = currPlayer == 1 ? player1Dice : player2Dice;
+  var playerNum;
```
lim-jiahao, line 70: nice use of ternary operators to simplify a vanilla if-else assignment. i personally prefer it too for simple cases like this but be aware that it can be unreadable for more complex cases, where a `if-else` would be more suitable. but good that you are using something that isn't taught, and it's one more tool at your disposal

script.js
```diff
@@ -1,4 +1,150 @@
...
+  else {
+    playerNum = Number(String(diceArray[1]) + String(diceArray[0]));
+  }
+
+  // Store player num in the relevant global player num variable
+  if (currPlayer == 1) {
+    player1Num = playerNum;
+  } else {
+    player2Num = playerNum;
+  }
```
lim-jiahao, line 85: this is so clean and smart. for simple if-else that only contain one liners in the body, you can actually do away with the curly braces. example lines 81 to 85 can be:

```
if (currPlayer == 1) player1Num = playerNum;
else player2Num = playerNum;
```
Some people swear by curly braces but I tend to judge it on a case-by-case basis depending on what I feel looks most organised and readable.

script.js
```diff
@@ -1,4 +1,150 @@
...
+ * Compute the winner between Player 1 and Player 2.
+ * Return either 1 or 2 to represent the winning player.
+ * In the event of a tie, Player 2 wins.
+ */
+var determineWinner = function () {
+  if (player1Num > player2Num) {
+    return 1;
+  }
+  return 2;
+};
```
lim-jiahao, line 101: nice use of early return + default return. since you used ternary operators earlier, let me show you how this can be reduced to one line by making use of them:

```
var determineWinner = function () {
  return player1Num > player2Num ? 1 : 2;
};

File: script.js [hunk 6]
```diff
@@ -1,4 +1,150 @@
...
+  if (gameMode == GAME_MODE_DICE_ROLL) {
+    // Get dice rolls for curr player and populate the curr player's dice array
+    var newDiceRolls = getDiceRolls();
+    // Switch mode to choose dice order
+    gameMode = GAME_MODE_CHOOSE_DICE_ORDER;
+    // Return the dice roll values to that player
+    return `Welcome Player ${currPlayer}. <br>
+      You rolled ${newDiceRolls[0]} and ${newDiceRolls[1]}. <br><br>
+      Please input 1 or 2 for your preferred order.<br>Option 1: ${newDiceRolls[0]}${newDiceRolls[1]}<br>Option 2: ${newDiceRolls[1]}${newDiceRolls[0]}`;
+  }
```
lim-jiahao, line 114: good

script.js
```diff
@@ -1,4 +1,150 @@
...
+
+  // Create a number based on the player's chosen dice order, and show it to the player
+  if (gameMode == GAME_MODE_CHOOSE_DICE_ORDER) {
+    // Validate the input. If first numeral index is neither 1 nor 2, tell the user.
+    var firstNumeralIndex = Number(input);
+    if (firstNumeralIndex != 1 && firstNumeralIndex != 2) {
+      return "Please input 1 or 2 for your preferred order.";
+    }
```
lim-jiahao, line 122: a point of improvement here:
you might wanna consider showing the dice rolls again in this message if they don't enter a valid 1 or 2. Because let's say I did not enter a 1 or 2 and end up getting this message, I might have forgotten already what my dice rolls are and need a reminder. Just another possible enhancement of user experience

lim-jiahao: also another point:
if the dice rolls are the same, does it make sense from a user experience pov to have the user choose the order? think about how we can skip this step and automatically compute the dice roll if they are the same

lim-jiahao:
Hey Olivia,

I am honestly super impressed. Your code is very clean and organised, and tbh it has been the most pleasant to review thus far. Your grasp of the concepts taught is very strong and you displayed it well, especially your logic in `getDiceRolls` and `getPlayerNumber` to store the dice rolls and score without having to repeat code unnecessarily. You also exhibited good understanding of early returns and the benefits it provides in not having to nest your code within many layers. This was something I emphasised a lot at the start so it makes me glad to see.

Some high level comments for improvement:
1. Try seeing if you can output a message into the output box immediately on page load by editing the html so that the user sees a message straight away after loading the game. Currently when I load your game, I am greeted with an empty text input and output box, and it is not immediately clear what I have to do. Always think about user experience when you are designing something!
2. Building on top of point 1, try making more html/css changes, maybe not for this proj but definitely for blackjack, even if it is simple things like changing the background color of the page or the container box, or changing the fonts that are being used. It's good practice and makes your project stand out. 
3. Try doing more than just Base for blackjack, I recognise that you might not have time but you are definitely more than capable of doing so and it would be a good way to challenge yourself. Excited to see what you come up with!

One thing - you definitely do not need to feel bad for referencing a lot. Programming is not a game of memory and I myself plus I'm sure a lot of way better developers still reference and Google a ton.

lim-jiahao:
On how to cater for more than 2 players, there are multiple ways to do so, but one I would recommend is to have an array `playerRolls` (this will be a 2D array) and `playerNums` which will respectively store each player's dice roll array and score, rather than the `player1Dice`, `player2Dice`, `player1Num`, and `player2Num` you currently have. You can imagine that doing it the current way would make it unscalable, especially if we have many players (imagine having 10 variables for player1Dice to player10Dice). When the game loads, ask the user for how many players are playing and push that number of empty arrays into `playerRolls` and that number of 0s into `playerNums`. Then change your game mode to `GAME_MODE_DICE_ROLL`. 

Your logic in `getDiceRolls`, `getPlayerNumber` and `determineWinner` would have to change to dynamically access/update the correct player's dice roll array and score based on the current player. I'll leave that for you to figure out! Similarly, the logic in line 130-142 to reset the curPlayer would have to change slightly.

PR #300, opened 2021-12-11
Basics B12S2 submission of beat that (wt)

weiting-ho:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

2-3 hours

N.A.

Trying to incorporate the More Comfortable portion

Comfort Level (1-5):
4

Completeness Level (1-5):
4

Not bad!

It runs as expected and it is pretty clear (at least to me).

N.A.

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,159 @@
-var main = function (input) {
-  var myOutputValue = 'hello world';
+/*
+There are 2 players and players take turns.
+When a player clicks Submit, the game rolls 2 dice and shows the dice rolls, for example 3 and 6.
+The player picks the order of the dice they want. For example, if they wanted the number 63, they would specify that the 2nd dice goes first. You can choose how the player specifies dice order.
+After both players have rolled and chosen dice order, the player with the higher combined number wins.
+
+How many game modes:
+1. Player 1 click submit to roll dice.
+2. Player 1 click 1 or 2 to choose order of dice. 
+3. Player 2 click submit to roll dice. 
+4. Player 2 click 1 or 2 to choose order of dice. 
+5. Determine who gets the higher combined number.  
+*/
+
+//global variables for current game mode
+var playerRoll1 = 0;
+var playerRoll2 = 0;
+var gameMode = "start game by rolling 2 dice";
+var currentPlayer = 1;
+var allDiceOrder = [];
+var accumulativeRound = 0;
+
+var playerRollDice = function () {};
+
+// random dice generator for user to roll dice
+var randomDiceGenerator = function () {
+  playerRoll1 = Math.floor(Math.random() * 6 + 1);
+  playerRoll2 = Math.floor(Math.random() * 6 + 1);
+  console.log(`dice1 output:`, playerRoll1, "& dice2 output:", playerRoll2);
+
+  if (playerRoll1 == playerRoll2) {
+    return `Welcome Player ${currentPlayer}.<br> <br> You have rolled ${playerRoll1} for both dice one and dice two. <br><br>Press 1.`;
+  } else
+    return `Welcome Player ${currentPlayer}. <br> <br> You have rolled ${playerRoll1} for dice one and ${playerRoll2} for dice two.<br> <br>Choose the order of the dice by entering "1" or "2".`;
+};
+
+// function for player to roll 2 dice to get 2 number
+var chooseDiceOrder = function (userInput) {
+  orderOfDice = userInput;
+  //input validation > error message for player if they did not enter 1 or 2
+  if (userInput != 1 && userInput != 2) {
+    return "Please enter either 1 or 2 to choose which dice to put infront.";
+  }
+
+  //if user choose 1 then first dice will be infront and second dice will be behind.
+
+  if (userInput == "1") {
+    var diceOrder = Number(String(playerRoll1) + String(playerRoll2));
+    console.log(`user chose ${playerRoll1}, then ${playerRoll2}.`);
+    //game mode switches to the next player
+    gameMode = "next player roll 2 dice";
+    console.log(gameMode);
+    myOutputValue = `Player ${currentPlayer}, you have chosen dice 1. <br><br>Your final value is ${playerRoll1}${playerRoll2}.<br><br> Please click submit.`;
+  }
+
+  //if user choose 1 then first dice will be infront and second dice will be behind.
+  if (userInput == "2") {
+    diceOrder = Number(String(playerRoll2) + String(playerRoll1));
+    console.log(`user chose ${playerRoll2}, then ${playerRoll1}.`);
+    gameMode = "next player roll 2 dice";
+    //game mode switches to the next player
+    console.log(gameMode);
+    myOutputValue = `You have chosen 2. <br><br>Your final value is ${playerRoll2}${playerRoll1}. <br> <br> Please click submit.`;
+  }
+
+  //to push the user's chosen number into an array and empty the temporary variables
+
+  if (accumulativeRound == 0) {
+    allDiceOrder.push(diceOrder);
+    console.log(allDiceOrder);
+  }
+
+  if (accumulativeRound >= 1) {
+    if (currentPlayer == 1) {
+      allDiceOrder[0] = allDiceOrder[0] + diceOrder;
+      console.log(`new array value = ${allDiceOrder[0]}`);
+      diceOrder = [];
+    }
+    if (currentPlayer == 2) {
+      allDiceOrder[1] = allDiceOrder[1] + diceOrder;
+      console.log(`new array value = ${allDiceOrder[0]}`);
+      diceOrder = [];
+    }
+    playerRoll1 = [];
+    playerRoll2 = [];
+  }
   return myOutputValue;
+};
+
+//functiont to determine the winner by comaring the values in the array
+var determineWinner = function () {
+  var myOutputValue = `Player 1 score is ${allDiceOrder[0]}<br>
+    Player 2 score is ${allDiceOrder[1]}<br>`;
+  var endOfSentence = `This is not the end! Click submit again to continue the game and try to beat your friend!  `;
+  if (allDiceOrder[0] >= allDiceOrder[1]) {
+    myOutputValue =
+      myOutputValue +
+      `<br><br><b>The winner is Player 1.</b><br><br>` +
+      endOfSentence;
+  }
+  if (allDiceOrder[1] >= allDiceOrder[0]) {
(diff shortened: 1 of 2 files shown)


script.js
```diff
@@ -1,4 +1,159 @@
...
+
+//global variables for current game mode
+var playerRoll1 = 0;
+var playerRoll2 = 0;
+var gameMode = "start game by rolling 2 dice";
+var currentPlayer = 1;
+var allDiceOrder = [];
+var accumulativeRound = 0;
+
+var playerRollDice = function () {};
```
lim-jiahao, line 23: what is this empty function doing here lol

script.js
```diff
@@ -1,4 +1,159 @@
...
+5. Determine who gets the higher combined number.  
+*/
+
+//global variables for current game mode
+var playerRoll1 = 0;
+var playerRoll2 = 0;
+var gameMode = "start game by rolling 2 dice";
+var currentPlayer = 1;
+var allDiceOrder = [];
+var accumulativeRound = 0;
```
lim-jiahao, line 21: nice and clear global variables. i would say best to have a comment on what `allDiceOrder` and `accumulativeRound` tracks, as it can be a bit unclear. alternatively, rename the variables to be something more descriptive then  the comment is unnecessary. this is another point for why descriptive variable names are desired as your code becomes self-documenting

script.js
```diff
@@ -1,4 +1,159 @@
...
+
+var playerRollDice = function () {};
+
+// random dice generator for user to roll dice
+var randomDiceGenerator = function () {
+  playerRoll1 = Math.floor(Math.random() * 6 + 1);
+  playerRoll2 = Math.floor(Math.random() * 6 + 1);
+  console.log(`dice1 output:`, playerRoll1, "& dice2 output:", playerRoll2);
```
lim-jiahao, line 29: generally ok but I would prob still abstract the ` Math.floor(Math.random() * 6 + 1)` part into a separate function called `rollDice` and just have `playerRoll1 = rollDice()` and `playerRoll2 = rollDice()`. two benefits: 1. this function becomes much more readable 2. if you have a bug with your dice roll generation, you only need to change the code in one place rather than two.

script.js
```diff
@@ -1,4 +1,159 @@
...
+
+  if (playerRoll1 == playerRoll2) {
+    return `Welcome Player ${currentPlayer}.<br> <br> You have rolled ${playerRoll1} for both dice one and dice two. <br><br>Press 1.`;
+  } else
+    return `Welcome Player ${currentPlayer}. <br> <br> You have rolled ${playerRoll1} for dice one and ${playerRoll2} for dice two.<br> <br>Choose the order of the dice by entering "1" or "2".`;
+};
```
lim-jiahao, line 35: ok too but would it be better if, when the dice rolls are the same, the user does not have to type anything and just has to press submit to move to the next step? 

because actually when i get similar dice rolls, if i enter something that isnt 1 or 2, i still get your message in line 42. not the best experience and would be ideal if the game auto calculates it for me if there is really only one outcome for the score

script.js
```diff
@@ -1,4 +1,159 @@
...
+
+  //if user choose 1 then first dice will be infront and second dice will be behind.
+  if (userInput == "2") {
+    diceOrder = Number(String(playerRoll2) + String(playerRoll1));
+    console.log(`user chose ${playerRoll2}, then ${playerRoll1}.`);
+    gameMode = "next player roll 2 dice";
+    //game mode switches to the next player
+    console.log(gameMode);
+    myOutputValue = `You have chosen 2. <br><br>Your final value is ${playerRoll2}${playerRoll1}. <br> <br> Please click submit.`;
+  }
```
lim-jiahao, line 64: this is fairly repetitive, notice the only difference between the two `if`s is really just line 48 and 58 so we prob can do that an `if-else`, and everything else is outside of it

```
  var diceOrder;
  if (userInput == "1") {
     diceOrder = Number(String(playerRoll1) + String(playerRoll2));
  } else if (userInput == "2") {
    diceOrder = Number(String(playerRoll2) + String(playerRoll1));
 }

    gameMode = "next player roll 2 dice";
    myOutputValue = `Player ${currentPlayer}, you have chosen dice ${userInput}. <br><br>Your final value is ${diceOrder}.<br><br> Please click submit.`;
```

lim-jiahao: the reason why i emphasise the point of repetition so much is because in programming, we try not to repeat ourselves if possible - the reason for this is 1. for readability purpose, and 2. for code maintainability purpose, imagine you notice a bug here and need to change something, then you need to change it twice for the code sections you repeated. And this gets more complex if there are more instances of unnecessary repetition in your code.

script.js
```diff
@@ -1,4 +1,159 @@
...
+    if (currentPlayer == 2) {
+      allDiceOrder[1] = allDiceOrder[1] + diceOrder;
+      console.log(`new array value = ${allDiceOrder[0]}`);
+      diceOrder = [];
+    }
+    playerRoll1 = [];
+    playerRoll2 = [];
+  }
+  return myOutputValue;
+};
```
lim-jiahao, line 88: a bit unclear what `allDiceOrder` was supposed to do until I reached here and realised its an array storing the cumulative score of both players, a better name might be `cumulativeScores`? but good work in trying out the more comfortable for keeping track of the score!

also, why is `playerRoll1` and `playerRoll2` reset to an empty array, it should be 0 right? doesn't really affect your code since you assign it back to a number value during the course of the next game but just curious if this was an intentional choice or just an oversight

script.js
```diff
@@ -1,4 +1,159 @@
...
+      `<br><br> <b>The winner is Player 2!</b><br><br>` +
+      endOfSentence;
+  }
+  if (allDiceOrder[1] == allDiceOrder[0]) {
+    myOutputValue =
+      myOutputValue + `<br><br> <b>It is a draw!</b><br><br>` + endOfSentence;
+  }
+
   return myOutputValue;
 };
```
lim-jiahao, line 113: this is good but should line 95 and 101 be a strictly more than condition? since in line 107 we are accounting for ties

script.js
```diff
@@ -1,4 +1,159 @@
...
+    //switch mode to choose dice order
+    gameMode = "player choose dice order";
+    console.log(gameMode);
+    return diceRoll;
+  }
+  if (gameMode == "player choose dice order") {
+    var diceScore = chooseDiceOrder(input);
+    return diceScore;
+  }
+};
```
lim-jiahao, line 130: good and clean function made possible by your earlier helper functions. one comment would be to make the `diceRoll` and `diceScore` variables more descriptive, from here given the names I would expect they are numbers but they are in fact strings

script.js
```diff
@@ -1,4 +1,159 @@
...
+
+var main = function (input) {
+  if (gameMode == "next player roll 2 dice") {
+    console.log(gameMode);
+    //loop to have more than 2 or as many players in the game.
```
lim-jiahao, line 135: does your game have this feature for more than 2 players? when playing around i only noticed a 2 player mode

script.js
```diff
@@ -1,4 +1,159 @@
...
+    while (currentPlayer < 2) {
+      currentPlayer += 1;
+      gameMode = "start game by rolling 2 dice";
+      console.log(gameMode);
+      return `It is now Player ${currentPlayer}'s turn! <br><br> Please click submit to roll the dice.`;
+    }
```
lim-jiahao, line 141: i think i get what you're trying to do here but i dont think a `while` loop is the most suitable here. because you return the moment the code enters one iteration of the loop, lines 137 to 140 is only going to run once each time the code comes into `main`, defeating the purpose of a loop! 

you probably want to do an `if` in line 136 paired together with an `else if` for the condition in line 143

script.js
```diff
@@ -1,4 +1,159 @@
...
+    console.log(gameMode);
+    var winner = determineWinner();
+    //game continues indefinitely
+    gameMode = "start game by rolling 2 dice";
+    currentPlayer = 1;
+    accumulativeRound = 1;
+    return winner;
+  }
+  return beatThatGame(input);
+};
```
lim-jiahao, line 159: i think your `main` and `beatThatGame` functions are sufficiently succinct that they can be merged together, this way the code becomes more readable too, as there is less jumping around between different functions and the single function will outline the game flow by nature of the order of the code

script.js
lim-jiahao: another point here: i noticed you have 4 game modes: "start game by rolling 2 dice", "player choose dice order", "next player roll 2 dice", and "determine winner among 2 players"

You typed out the string values of these game modes explicitly in your `beatThatGame` and `main` function, a better practice would be to define them as global variables and in your functions, reference them using the variables when you need to check for what the game mode is. To illustrate:

```
const START_GAME = "start game by rolling 2 dice";
const NEXT_PLAYER = ""next player roll 2 dice";
const CHOOSE_ORDER = "player choose dice order";
const DETERMINE_WINNER = "determine winner among 2 players";
var gameMode = START_GAME;
```
This goes together with your other global variables here, then for example in line 133 when you check `if (gameMode == "next player roll 2 dice")` you can do `if (gameMode == NEXT_PLAYER)` instead. Likewise, when you are updating the game mode in example line 122, `gameMode = "player choose dice order"` you do `gameMode = CHOOSE_ORDER;` instead.

This reduces code repetition as you only need to use the string value once which prevents errors arising from typos. It also lets me know at a high level glance just by looking at the start of your code, what state variables to expect. I am using the `const` keyword to declare them rather than var. For constants, some people will use SCREAM_CASE as well for the variable name. I know this wasn't really covered in the course but it's a minor step up if you're interested. 

For more info:

lim-jiahao:
Hey Wei Ting,

Generally very solid code. A lot of my comments really are just about reducing code repetition and suggestions on how to refactor. You demonstrate good understanding of the concepts so far and put it to good use! You are right in that your code is for the most part clear, but do insert more comments especially for those sections that are slightly more complex. 

High level comments:
- Variable naming can be clearer
- Refactor to reduce code repetition and enhance readability + maintainability
- More comments at appropriate places

I am very pleased with your attempt at making changes to the HTML and CSS, especially as you are interested in bootcamp. As I mentioned, the first part of bootcamp is mainly front-end doing DOM manipulation using HTML/CSS and JavaScript so it would be good to have some exposure right now, even if it is simple things like changing the background color of the page or the container box, or changing the fonts that are being used. Do keep it up for blackjack! 

Nice attempt at some of the more comfortable portions too. Did you really do all this in 2-3 hours? thats crazy fast

Hope to see more of this or better in blackjack. Excited to see what you come up with!

PR #304, opened 2021-12-11
Project submission for beatThat

krystalng1:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

How many hours did you spend on this assignment? // 5 Hours 

// Not unable to pull the helper function after refactoring to change the player and choose the dice order

// To get the logic to change the player and and choose the dice order and afterwards refactor the codes. 

Comfort Level (1-5): // 3

Completeness Level (1-5):  // 3

Is there anything in this code that you feel pleased about? //Able to run the code eventually, setting up the various helper factor so that i can use it in the main function itself and successfully managed to refactor the codes

What's one aspect of your code you would like specific, elaborate feedback on? // Managed to understand how to use the various game mode to change the player and choose the dice order.

### DIFF
--- a/script.js
+++ b/script.js
@@ -9,7 +9,7 @@
 
 // 4. After both players have rolled and chosen dice order, the player with the higher combined number wins. -> to compare the 2 dice value
 
-// 6. The game will reset and will go back to Player 1
+// 5. The game will reset and will go back to Player 1
 
 //To define Global Variable
 var gameModeRollDice = "game mode roll dice";
--- a/script.js
+++ b/script.js
@@ -1,4 +1,135 @@
+// Project 2 : Beat that (Base)//
+
+// 1. Number of players = 2 , will take turns to play the game
+// Create a prompt to inform player to click on submit button.
+
+// 2. When a player clicks Submit, the game rolls 2 dice and shows the dice rolls, for example 3 and 6. The dice values should be stored for each player.
+
+// 3. The player picks the order of the dice they want. (e.g whether dice 1 or dice 2 will go first) -> conditional statement. Store the value in an array
+
+// 4. After both players have rolled and chosen dice order, the player with the higher combined number wins. -> to compare the 2 dice value
+
+// 6. The game will reset and will go back to Player 1
+
+//To define Global Variable
+var gameModeRollDice = "game mode roll dice";
+var gameModeChooseDice = "game mode choose dice";
+var gameModeDetermineWinner = "game mode determine winner";
+//start the gamemode to roll the dice
+var gameMode = gameModeRollDice;
+var currentPlayerRoll = [];
+//To track the current player along the game
+var currPlayer = 1;
+var allPlayerScore = [];
+
+//Roll dice helper function to return random value from 1 - 6
+var rollDice = function () {
+  var randomDecimal = Math.random() * 6;
+  var randomInteger = Math.floor(randomDecimal);
+  var diceNumber = randomInteger + 1;
+  return diceNumber;
+};
+
+// Roll 2 dice helping function to return message
+var rollDiceForGame = function () {
+  var counter = 0;
+  while (counter < 2) {
+    currentPlayerRoll.push(rollDice());
+    counter = counter + 1;
+  }
+  console.log("rolldice for player", currentPlayerRoll);
+
+  return `🎲 Welcome Player ${currPlayer} 🎲 <br><br> You rolled ${currentPlayerRoll[0]} for Dice 1 and ${currentPlayerRoll[1]} for Dice 2 <br><br> Choose the order of the dice by entering the number "1" or "2".`;
+};
+
+var getPlayerScore = function (playerInput) {
+  var playerScore;
+  console.log("playerScore");
+  // to prompt error message if the input is not '1' or '2'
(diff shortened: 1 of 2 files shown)


script.js
```diff
@@ -1,4 +1,135 @@
...
+
+// 4. After both players have rolled and chosen dice order, the player with the higher combined number wins. -> to compare the 2 dice value
+
+// 5. The game will reset and will go back to Player 1
+
+//To define Global Variable
+var gameModeRollDice = "game mode roll dice";
+var gameModeChooseDice = "game mode choose dice";
+var gameModeDetermineWinner = "game mode determine winner";
+//start the gamemode to roll the dice
```
kenrickles, line 18: I think a clearer comment here would be 'initialising the game mode to roll the dice'

script.js
```diff
@@ -1,4 +1,135 @@
...
+    );
+  }
+  allPlayerScore.push(playerScore);
+  currentPlayerRoll = [];
+  return `🎲 Player ${currPlayer}🎲 <br><br>
+  Your choosen value is ${playerScore}.`;
+};
+
+var determineWinner = function () {
+  var winnerMessage = `🎲 Player 1 score is ${allPlayerScore[0]} and Player 2 score is ${allPlayerScore[1]}🎲  <br>`;
```
kenrickles, line 73: Good use of defining variables that will be repeated !

script.js
```diff
@@ -1,4 +1,135 @@
...
+  console.log("check game mode", gameMode);
+  var outputMessage = "";
+  // If the game mode is roll dice, we will call the helper function and generate the dice number for the player
+  if (gameMode == gameModeRollDice) {
+    outputMessage = rollDiceForGame();
+    gameMode = gameModeChooseDice;
+    return outputMessage;
+  }
+  // If game mode changes to choose dice, we will run this section of codes
+  console.log("check game mode change", gameMode);
```
kenrickles, line 110: i would actually shift the console.log into line 112 instead. to help me better see if the change happened and in a way to track the change.

script.js
```diff
@@ -1,4 +1,135 @@
...
+  if (gameMode == gameModeChooseDice) {
+    outputMessage = getPlayerScore(input);
+
+    // if this is still player 1, we will run this section and inform that it is now playaer 2 turn to roll the dice
+    console.log("curr player", currPlayer);
```
kenrickles, line 115: i will shift this console.log into line 117, to check for the current player.

script.js
```diff
@@ -1,4 +1,135 @@
...
+    if (currPlayer == 1) {
+      currPlayer = 2;
+      gameMode = gameModeRollDice;
+      return `${outputMessage} It is Player 2's turn after you have selected the dice order. <br>`;
+    }
+    // if this is player 2, we will switch the game mode to determine the winner for both players
+    console.log("curr player", currPlayer);
```
kenrickles, line 122: Same as the above comment, but i will shift it in line 124 instead

script.js
```diff
@@ -1,4 +1,135 @@
...
+      return `${outputMessage} <br><br> Please press submit to determine the winner!`;
+    }
+  }
+  // call the helper function above to determine the winner and restart the game afterwards
+  if (gameMode == gameModeDetermineWinner) {
+    outputMessage = determineWinner();
+    restartGame();
+    return outputMessage;
+  }
+  console.log("Game is restart");
```
kenrickles, line 134: i would shift this to line 133 instead.

script.js
```diff
@@ -1,4 +1,135 @@
...
+};
+
+var getPlayerScore = function (playerInput) {
+  var playerScore;
+  console.log("playerScore");
+  // to prompt error message if the input is not '1' or '2'
+  if (playerInput != 1 && playerInput != 2) {
+    return `Please only input either '1' or '2' to choose which dice to use as the first digit.`;
+  }
+  console.log("input==1");
```
kenrickles, line 52: i would put this inside the if function and maybe write "Running in If condition, PlayerInput = 1"

script.js
```diff
@@ -1,4 +1,135 @@
...
+  // if the input is '1', they will recognise the player score in this order
+  if (playerInput == 1) {
+    playerScore = Number(
+      String(currentPlayerRoll[0]) + String(currentPlayerRoll[1])
+    );
+  }
+  console.log("input==2");
```
kenrickles, line 59: Same as the above comment

kenrickles (commented):
Excellent effort! i really see the thought being put into this. I really like how you try to use console.log. I like your comments, it really helps me to follow along easily!


A point to note: When we use console.log, we should try to put it inside the IF function itself as it will tell us where the code is actually running. So if something doesn't work right, we can check if it enters the IF condition or not easier. I would prefer if you could make the main function cleaner for your final project. 

Also, I would like you to think about the user's perspective more. Perhaps you could get a friend or a family member to try playing your game to give you feedback on what the user's painpoints are.

 For example, when you return the message: "Your choosen value is 12. It is Player 2's turn after you have selected the dice order."  The user actually has no idea that he/she has to press the submit button again to move on to next player. 
 
 Remember, the reason we become software engineers is to make the user experience great as well. :) 
 
 Overall, excellent effort and good work!!

PR #311, opened 2021-12-13
beat that submission 12-1 Ma he

mahe993:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

1 day

too many or not enough "}"

trying to implement knockout mode in a round robin tournament format. still trying to work on it.

Comfort Level (1-5): 1 

Completeness Level (1-5): 2.5

What did you think of this deliverable? needed more time and review on this

Is there anything in this code that you feel pleased about? n.a.

What's one aspect of your code you would like specific, elaborate feedback on? if there are any glaring errors up till HIGHSCORE mode.

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,285 @@
+//states
+var playerCount = 0;
+var userNames = [];
+var gameMode = "";
+var gameType = "";
+var diceCount = 0;
+var playerScores = [];
+var diceRolls = [];
+var playGameCount = 0;
+var scoreBoard = [];
+var tempScoreBoard = [];
+
+//constants
+const NORMAL = "Normal";
+const LOWEST = "Lowest";
+const KNOCKOUT = "Knockout";
+const HIGHSCORE = "Highscore";
+const DICE = [1, 2, 3, 4, 5, 6];
+
+//username change state
+var userNameCount = 11;
+
 var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
+  // player count validation
+  if (playerCount == 0) {
+    if (isNaN(input) || input > 10 || input < 2) {
+      return `Please enter the number of players that wish to play. Maximum of 10 and minimum of 2!`;
+    }
+    playerCount = input;
+    // update scoreboard
+    for (let a = 0; scoreBoard.length < playerCount; a += 1) {
+      scoreBoard.push(Number(0));
+    }
+
+    console.log("scoreboard fill 0: " + scoreBoard);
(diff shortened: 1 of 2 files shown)


script.js
```diff
@@ -1,4 +1,443 @@
...
+var playerScores = [];
+var diceRolls = [];
+var playGameCount = 0;
+var scoreBoard = [];
+var tempScoreBoard = [];
+var gameRoundCount = 0;
+var eliminationArray = [];
+
+//constants
+const NORMAL = "Normal";
```
jeremylim91, line 16: Good usage of string constants!

script.js
```diff
@@ -1,4 +1,443 @@
...
+    }
+    diceCount = Number(input);
+    return `<b>${diceCount}</b> dices selected. Please proceed to select your game mode.<br><br>Modes available:<br>⮞Highscore<br>⮞Knockout`;
+  }
+
+  //gameMode validation
+  if (gameMode == "") {
+    if (
+      input == "" ||
+      (input.toUpperCase() != HIGHSCORE.toUpperCase() &&
```
jeremylim91, line 75: Good practice of changing the case of the input so that user errors can't break your code. Can you think of how we can transform input into uppercase in just one line, rather than every instance that we use input? (e.g. in line 76)

script.js
```diff
@@ -1,4 +1,443 @@
...
+        }</b> rolled ${diceRolls}, resulting in a score of ${
+          tempScoreBoard[playGameCount - 1]
+        }.<br><br><b>${
+          userNames[playGameCount]
+        }</b> click submit to roll. ${(diceRolls = [])}`;
+      }
+    }
+  }
+  //KNOCKOUT game mode
+  else if (gameMode == KNOCKOUT.toUpperCase()) {
```
jeremylim91, line 272: In this case, actually we could've just used 'KOCKOUT' (i.e. all caps) string in line 18. Then we wouldn't have to transform it to upper.

jeremylim91 (commented):
Hey @mahe993 , good job on this code, and on pushing yourself to change the CSS/appearance of the UI. I think it's clear that you are able to handle complex code logic; sincej its clear you can alr code, one area to focus on is on adopting best practices, and based on your code, I think the specific area to look at would be refactoring code to abstract implementation details out of main. 

Good job once again, you should be really proud of this code! :) Do deploy it so that your friends can play it as well.

script.js
mahe993: Thank you for taking the time to review! really appreciate it!

PR #315, opened 2021-12-14
12-4 Shak's Beat That

supashak:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

2-3 hours

undefined because I forgot to set the parameters in the helper functions I created

The base. 

Comfort Level (1-5):
3

Completeness Level (1-5):
3

I would have liked to spend more time on the project but I have exams right now.

Possibly on refactoring...I'm not too sure if I have refactored all that I could have refactored.

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,73 @@
+var playerOneGuess = "";
+var playerTwoGuess = "";
+var diceRollOne = 0;
+var diceRollTwo = 0;
+var playerOneNumber = 0;
+var playerTwoNumber = 0;
+var gameState = "";
+var myOutputValue = "";
+
+//dice roll function
+var diceRoll = function () {
+  var oneToFive = Math.floor(Math.random() * 6);
+  var randomNumber = oneToFive + 1;
+  return randomNumber;
+};
+//Helper function for selecting combination which gives largest value
+var playerOneCombinationGenerator = function () {
+  diceRollOne = diceRoll();
+  diceRollTwo = diceRoll();
+  if (diceRollOne >= diceRollTwo) {
+    var playerOneGuess = `${diceRollOne}${diceRollTwo}`;
+    playerOneNumber = Number(playerOneGuess);
+    myOutputValue = `Welcome Player One! You have rolled ${diceRollOne} for Dice One and ${diceRollTwo} for Dice Two. <br> The largest possible combination is ${playerOneGuess}.It is now Player 2\'s turn.`;
+  }
+  if (diceRollTwo > diceRollOne) {
+    var playerOneGuess = `${diceRollTwo}${diceRollOne}`;
+    playerOneNumber = Number(playerOneGuess);
+    myOutputValue = `Welcome Player One! <br> You have rolled ${diceRollOne} for Dice One and ${diceRollTwo} for Dice Two. <br> The largest possible combination is ${playerOneGuess}.It is now Player 2\'s turn.`;
+  }
+  return myOutputValue;
+};
+
+var playerTwoCombinationGenerator = function () {
+  diceRollOne = diceRoll();
+  diceRollTwo = diceRoll();
+  if (diceRollOne >= diceRollTwo) {
+    var playerTwoGuess = `${diceRollOne}${diceRollTwo}`;
+    playerTwoNumber = Number(playerTwoGuess);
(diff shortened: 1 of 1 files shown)


script.js
```diff
@@ -1,4 +1,73 @@
+var playerOneGuess = "";
+var playerTwoGuess = "";
+var diceRollOne = 0;
+var diceRollTwo = 0;
+var playerOneNumber = 0;
+var playerTwoNumber = 0;
+var gameState = "";
+var myOutputValue = "";
```
shannonssd, line 8: Good placement, naming and use of global variables

script.js
```diff
@@ -1,4 +1,73 @@
...
+  diceRollTwo = diceRoll();
+  if (diceRollOne >= diceRollTwo) {
+    var playerOneGuess = `${diceRollOne}${diceRollTwo}`;
+    playerOneNumber = Number(playerOneGuess);
+    myOutputValue = `Welcome Player One! You have rolled ${diceRollOne} for Dice One and ${diceRollTwo} for Dice Two. <br> The largest possible combination is ${playerOneGuess}.It is now Player 2\'s turn.`;
+  }
+  if (diceRollTwo > diceRollOne) {
+    var playerOneGuess = `${diceRollTwo}${diceRollOne}`;
+    playerOneNumber = Number(playerOneGuess);
+    myOutputValue = `Welcome Player One! <br> You have rolled ${diceRollOne} for Dice One and ${diceRollTwo} for Dice Two. <br> The largest possible combination is ${playerOneGuess}.It is now Player 2\'s turn.`;
```
shannonssd, line 28: We typically use another variable instead of myOutputValue in the helper functions, so that we can use myOutputValue in the main function

script.js
```diff
@@ -1,4 +1,73 @@
...
+    myOutputValue = `Welcome Player Two! <br> You have rolled ${diceRollOne} for Dice One and ${diceRollTwo} for Dice Two. <br> The largest possible combination is ${playerTwoGuess}.It is now time to decide the winner.`;
+  }
+  return myOutputValue;
+};
+
 var main = function (input) {
-  var myOutputValue = 'hello world';
+  if (input == "player 1") {
+    gameState = "player 2";
+    playerOneCombinationGenerator();
```
shannonssd, line 52: It would be more consistent to: 
- execute playerOneCombinationGenerator(); 
- then assign it to a variable,
- then return the variable

For e.g.: var myOutputValue =  playerOneCombinationGenerator();
return myOutputValue

script.js
```diff
@@ -1,4 +1,73 @@
...
+//dice roll function
+var diceRoll = function () {
+  var oneToFive = Math.floor(Math.random() * 6);
+  var randomNumber = oneToFive + 1;
+  return randomNumber;
+};
+//Helper function for selecting combination which gives largest value
+var playerOneCombinationGenerator = function () {
+  diceRollOne = diceRoll();
+  diceRollTwo = diceRoll();
```
shannonssd, line 19: This could be in its own helper function to make playerOneCombinationGenerator() shorter and more concise

shannonssd (approved):
Great job overall Shak!

Just remember that we would like to execute a function and store its value into a variable and then return the variable, rather than just executing the function by itself.

Otherwise, keep up the great work! 😊

script.js
supashak: Ah ok...so for the helper function I should just have ended it at line 27 to generate the highest possible number?

script.js
supashak: I'm so sorry but I'm not too sure what you mean by this. What would be the purpose of the new helper function? To just generate two dice rolls?

script.js
shannonssd: Nope the code is correct. The point is that different functions should not be returning the same variable as this can lead to errors when the computer is reading the programme. 

So try to use and return different variables in each function instead of myOutputValue at lines 30, 46, 72.
We don't typically make myOutputValue a global variable so if you need to then something isn't quite right

script.js
shannonssd: This may not be the best example since it's just two lines, but typically when there is code that repeats, we want to put them into a helper function. 

One basic coding principle is to keep your code DRY: Don't Repeat Yourself. And since you're gonna roll two dice at lines 18-19 and again at lines 34-35, we would rather put it into a helper function and call the helper function here.

Imagine having 10 players instead of just 2? That would be an additional 10 lines of code. Now image if there was three other areas that you could have refactored into a helper function? The code would have more lines than necessary and would be less readable than it could potentially be.

Hope this helps!

PR #332, opened 2022-02-22
Batch 13-5 Nicola Beat That

nicolalew:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

How many hours did you spend on this assignment? 5

Got stuck in my input validation loop until I remembered that the input is a string. Had to change to if input == "1" rather than if input == 1 to escape the input validation loop.

Comfort Level (1-5): 3

Completeness Level (1-5): 3

It was interesting thinking about how to design the logic to allow room for expansion e.g. using an array for player scores or using loops to allow for more than 2 players in the future.

Use of arrays and a simple while loop.

How can I return a message telling the player what their score is after they choose an order while still in my while loop? Or should I use a different loop?

### DIFF
--- a/script.js
+++ b/script.js
@@ -10,6 +10,10 @@ var playerScores = [];
 var playerRolls = [];
 var myOutputValue = "";
 var playerTurn = 1;
+var numPlayers = 0;
+
+// possible gameModes
+// "dice roll", "choose order"
 
 // create function that rolls a random dice number
 var rollDice = function () {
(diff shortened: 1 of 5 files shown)


script.js
```diff
@@ -26,40 +30,50 @@ var rollDice = function () {
 var playerRoll = function () {
   playerRolls.push(rollDice());
   playerRolls.push(rollDice());
+  console.log(`playerRolls: ${playerRolls}`);
   myOutputValue = `Welcome Player ${playerTurn}. <br> Dice 1: ${playerRolls[0]} <br> Dice 2: ${playerRolls[1]} <br> Please choose the order of the dice.`;
   return myOutputValue;
 };
 
 var main = function (input) {
```
Cheoklate, line 38: I would recommend starting to move a majority of your code outside the var main function, and instead create a bunch of helper functions outside the var main.  For example in this situation, you could move the majority of the code outside the var main and have a function that encompasses all the game logic.
```
var gameLogic= function (input) {
... // code here
}

var main = function (input) {
  return gameLogic(input);
};
```
Helps with readability and if you ever want to use the gameLogic outside of the var main function you'll be able to do that.

Cheoklate (commented):
Nice clean and concise. Demonstrates a good understanding of arrays and conditionals. 

- Good job editing the HTML even though that has yet to be taught!
- Nice use of input validation

Cheoklate (commented):
Dissmiss comment

script.js
```diff
@@ -1,4 +1,105 @@
...
+  var diceNumber = randomInteger + 1;
+  return diceNumber;
+};
+
+// create function that rolls 2 dice for player, stores these numbers in an array and returns a message informing player of the 2 dice numbers
+var playerRoll = function () {
+  playerRolls.push(rollDice());
+  playerRolls.push(rollDice());
+  console.log(`playerRolls: ${playerRolls}`);
+  myOutputValue = `Welcome Player ${playerTurn}. <br> Dice 1: ${playerRolls[0]} <br> Dice 2: ${playerRolls[1]} <br> Please choose the order of the dice.`;
```
Cheoklate, line 34: Instead of defining my OutputValue then returning it you could instead just return the string straight away.
```suggestion
  return `Welcome Player ${playerTurn}. <br> Dice 1: ${playerRolls[0]} <br> Dice 2: ${playerRolls[1]} <br> Please choose the order of the dice.`;
```
Tidies up code and removes unnecessary variables.

PR #335, opened 2022-02-22
basics-beat-that_ShiChie

JakeChing:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

How many hours did you spend on this assignment? 5-6 hours

Please fill in one error and/or error message you received while working on this assignment. Keep getting undefined. But managed to find the error while using console.log

What part of the assignment did you spend the most time on? Adding player 2 into the game.

Comfort Level (1-5): 3

Completeness Level (1-5): 4

What did you think of this deliverable? Good. It covers from topic from day 1 to the last class. 

Is there anything in this code that you feel pleased about? I wanted to add the names for both players. It worked all the way until I have to use "player 1 & 2" at the end. Else it will be more complex and longer. 

What's one aspect of your code you would like specific, elaborate feedback on? Nope. I do find that using helper function is much neater than putting all the function in the main function.

### DIFF
--- a/script.js
+++ b/script.js
@@ -33,7 +33,7 @@ var rollDiceForPlayer = function () {
 
 var getPlayerScore = function (playerInput) {
   // input validation
-  if (playerInput != 1 && input != 2) {
+  if (playerInput != 1 && playerInput != 2) {
     return `Error! Please only input "1" or "2" to choose your first digit of your final value. <br> Dice 1: ${currentPlayerRolls[0]} | Dice 2: ${currentPlayerRolls[1]}`;
   }
 
@@ -52,7 +52,7 @@ var getPlayerScore = function (playerInput) {
   }
   // store current player score
   allPlayerSCore.push(playerFinalValue);
-
+  console.log(playerFinalValue);
   // clear current player rolls array
   currentPlayerRolls = [];
   return `${userName}, your final value is ${playerFinalValue}.`;
@@ -75,15 +75,24 @@ var main = function (input) {
     // call getPlayerScore helper function
     myOutputValue = getPlayerScore(input);
 
-    if (currentPlayer == 1){
+    if (currentPlayer == 1) {
       currentPlayer = 2;
       currentGameMode = gameModeDiceRoll;
+      console.log(currentPlayer);
       return `${myOutputValue} <br> It is now player 2's turn. Please input your name!`;
     }
 
-    if (currentPlayer == 2){
-      currentGameMode = gameCompareScore
-      
-      return `${myOutputValue} <br><br> Please click submit to calculate score`
+    if (currentPlayer == 2) {
+      currentGameMode = gameCompareScore;
+      return `${myOutputValue} <br><br> Please click submit to calculate score`;
     }
+  }
+
+  // if (currentGameMode == gameCompareScore){
+  //   myOutputValue = ``
+  //   // player 1 wins
+
+  //   //player 2 wins
+
+  //   //tie
 };
--- a/script.js
+++ b/script.js
@@ -1,4 +1,60 @@
+var gameModeDiceRoll = "game_state_dice_roll";
+var gameStateChooseDiceOrder = "game_state_choose_dice";
+var currentGameMode = "game_state_dice_roll";
+var userName = "";
+
+//create an array to store the values of the 2 dices
+var playerRolls = [];
+
+//creating helper function for dice roll
+var rollDice = function () {
+  var randomDecimal = Math.random() * 6;
+  var randomInteger = Math.floor(randomDecimal);
+  var diceNumber = randomInteger + 1;
+  return diceNumber;
+};
+
+//creating helper function to roll 2 dice
+var rollDiceForPlayer = function () {
+  var counter = 0;
+  while (counter < 2) {
+    playerRolls.push(rollDice());
+    counter = counter + 1;
+  }
+  //after this step, ask player to choose the order
+  return `Welcome ${userName}! <br> You rolled ${playerRolls[0]} for Dice 1 and ${playerRolls[1]} for Dice 2. <br> Please choose the dice number (either 1 or 2) to be your first digit of your final value.`;
+};
+
 var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
+  if (currentGameMode == gameModeDiceRoll) {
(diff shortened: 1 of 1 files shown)


script.js
```diff
@@ -1,4 +1,123 @@
+var gameModeDiceRoll = "game_state_dice_roll";
+var gameStateChooseDiceOrder = "game_state_choose_dice";
+var gameCompareScore = "game_compare_score";
+var currentGameMode = gameModeDiceRoll;
+
+var currentPlayer = 1;
+// var userName = "";
+var allPlayerSCore = [];
```
lim-jiahao, line 8: nice use of game modes and global vars

script.js
```diff
@@ -1,4 +1,123 @@
...
+//create an array to store the values of the 2 dices for both players
+var currentPlayerRolls = [];
+
+//creating helper function for dice roll
+var rollDice = function () {
+  var randomDecimal = Math.random() * 6;
+  var randomInteger = Math.floor(randomDecimal);
+  var diceNumber = randomInteger + 1;
+  return diceNumber;
+};
```
lim-jiahao, line 19: this was shown step by step in the preclass to simplify and make it easier to understand but consider condensing line 15-18 into one line in your code

return Math.floor(Math.random() * 6) + 1

script.js
```diff
@@ -1,4 +1,123 @@
...
+//creating helper function to roll 2 dice
+var rollDiceForPlayer = function () {
+  var counter = 0;
+  while (counter < 2) {
+    currentPlayerRolls.push(rollDice());
+    counter = counter + 1;
+  }
+  //after this step, ask player to choose the order
+  console.log(currentPlayerRolls[0]);
+  console.log(currentPlayerRolls[1]);
```
lim-jiahao, line 30: good use of console.log statements for debugging but they are usually removed before shipping to production in a professional setting (the equivalent of submitting your project for this course)

applies for your other console.log statements too

script.js
```diff
@@ -1,4 +1,123 @@
...
+      String(currentPlayerRolls[0]) + String(currentPlayerRolls[1])
+    );
+  }
+
+  // input == 2
+  if (playerInput == 2) {
+    var playerFinalValue = Number(
+      String(currentPlayerRolls[1]) + String(currentPlayerRolls[0])
+    );
+  }
```
lim-jiahao, line 52: since we have the above input validation, we do not need to check explicitly for both cases. we can just have a default case and check for the other case

```
var playerFinalValue = Number(
      String(currentPlayerRolls[0]) + String(currentPlayerRolls[1]);

if (playerInput == 2) {
   playerFinalValue = Number(
      String(currentPlayerRolls[1]) + String(currentPlayerRolls[0])
    );
  }
```

script.js
```diff
@@ -1,4 +1,123 @@
...
+  //player 2 wins
+  if (allPlayerSCore[0] < allPlayerSCore[1]) {
+    compareOutputValue = `${compareOutputValue} <br><br> Player 2 wins!`;
+  }
+  //tie
+  if (allPlayerSCore[0] == allPlayerSCore[1]) {
+    compareOutputValue = `${compareOutputValue} <br><br> It's a tie!`;
+  }
+  return compareOutputValue;
+};
```
lim-jiahao, line 78: consider refactoring this to use either early returns or a default value so that we only need to explicitly check for 2 out of 3 of the scenarios rather than all

script.js
```diff
@@ -1,4 +1,123 @@
...
+  console.log(currentPlayer);
+  var myOutputValue = "";
+
+  if (currentGameMode == gameModeDiceRoll) {
+    // set the name
+    // userName = input;
+    // now that we have the name for player 1 and rolled 2 dices, switch the mode
+    currentGameMode = gameStateChooseDiceOrder;
+    console.log(currentGameMode);
+    return (myOutputValue = rollDiceForPlayer());
```
lim-jiahao, line 97: this is not necessary, can just do `return rollDiceForPlayer();`

what will happen is that rollDiceForPlayer will be called, all the logic in it will run, then it returns a string

this line will then resolve to `return <whatever string was returned>`

script.js
```diff
@@ -1,4 +1,123 @@
...
+  }
+
+  if (currentGameMode == gameStateChooseDiceOrder) {
+    // call getPlayerScore helper function
+    myOutputValue = getPlayerScore(input);
```
lim-jiahao, line 102: theres a bug in this line

image

your input validation for `getPlayerScore` will return out of `getPlayerScore` with your input validation msg but after it is returned to `main`, the `main` function doesn't stop running because you didn't specify it to

thats why in the above screenshot, the gameplay still moves on to player 2

lim-jiahao: then when the gameplay moves on to player 2 and i play through it properly, player 2's score will be stored as player 1's score and i get this eventual result 

image

lim-jiahao: consider how you can change this up so that the main function stops running here if the input is not valid (hint: maybe shift the input validation out of `getPlayerScore` and into `main`, then return out of `main` with the relevant msg if the input is not valid)

script.js
```diff
@@ -1,4 +1,123 @@
...
+      currentGameMode = gameModeDiceRoll;
+      console.log(currentPlayer);
+      return `${myOutputValue} <br> It is now player 2's turn. Click submit when ready!`;
+    }
+
+    if (currentPlayer == 2) {
+      currentGameMode = gameCompareScore;
+      return `${myOutputValue} <br><br> Please click submit to calculate score`;
+    }
+  }
```
lim-jiahao, line 115: nice logic here to decide how to progress with the game depending on whose turn it currently is

lim-jiahao:
hi Shi Chie,

overall very good work! i can see the improvement from your SPS project esp in aspects like code cleanliness and less code repetition so keep it up for blackjack!

my comment from previously will still apply:
try playing around with the HTML (and CSS if you have time) for your next project. Currently when I play your game, I am greeted with an empty output box and it is not immediately clear what I have to do. Always take user experience into account when you code. A good way to improve on this is to simply edit the HTML so that a message is in the output box immediately when the game is loaded, perhaps to welcome the user and prompt the user to enter their name or give instructions on the rules of the game, how the game is played

Also try the more comfortable portions too! Last project already so push yourself more, you can do it 👍

PR #340, opened 2022-02-22
Basics 13-5 Cherylyn

cherylyn7:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

**How many hours did you spend on this assignment?**
8
**Please fill in one error and/or error message you received while working on this assignment.**
Putting all the different conditions  etc for the loops/nested loops at the right place

**What part of the assignment did you spend the most time on?**
Figuring out the right flow and logic for everything

**Comfort Level (1-5):**
2

**Completeness Level (1-5):**
4

**What did you think of this deliverable?**
I enjoyed this project. I was too ambitious in beginning of the project and wanted to code many options. But then I ran into issues with putting everything together in time, so went back to the drawing board to focus on the base.

**Is there anything in this code that you feel pleased about?**
Figuring out the nested loops and automating the play process.

**What's one aspect of your code you would like specific, elaborate feedback on?**
I'm worried that my basic understanding of concepts is wrong/I'm doing things the long way. So please let me know if there's an easier way to do things!

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,283 @@
+//Game Rules:
+//This game has a few modes:
+//Variables that players can adjust: Number of dice, number of players, win condition, playmode
+//Mode 1A: Highest number concatenated
+//Mode 1B: Highest number added together
+//Mode 1C: Highest number multipled
+//Mode 2A: Lowest number concatenated
+//Mode 2B: Lowest number added together
+//Mode 2C: Lowest number multipled
+
+//Game Settings input by Player
+var numberOfPlayers = ``;
+var numberOfDice = ``;
+var numberOfRounds = ``;
+var numString = [`zero`, `one`, `two`, `three`, `four`, `five`];
+
+//Playing Process
+var playersNames = [];
+var playersDiceRolls = [];
+var playersSubmissions = [];
+
+// Win-Loss Records
+var winLossRecords = [];
+
+// Helper Functions
+
+//Initial Game State
+var gameState = `summonTheGamemaster`;
+console.log(`gameState`, gameState);
+var roundsPlayed = 0;
+
+// Main Function
 var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
+  input = input.toLowerCase();
+
+  if (numString.includes(input) == true) {
+    input = numString.indexOf(input);
+  }
+
+  //Game settings code begins here
+
+  if (
+    gameState == `summonTheGamemaster` &&
+    input != `open the chamber of secrets`
+  ) {
+    return `The Gamemaster is quite a particular character. Say the words he wants or he will not entertain you.`;
+  }
+
+  if (
+    gameState == `summonTheGamemaster` &&
+    input == `open the chamber of secrets`
+  ) {
+    gameState = `gameSettingsNumPlayers`;
+    console.log(`gameState`, gameState);
+    return `Ah...well hello there. I see you have stumbed upon this little corner of the internet.<br>
+    <br>
+Perhaps you'd like to play a game? I promise it'll be fun... for me at least.<br>
+<br>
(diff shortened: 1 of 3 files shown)


script.js
```diff
@@ -1,4 +1,343 @@
...
+
+// Win-Loss Records
+var winLossRecords = [];
+var checkTiedRolls = [];
+var checkTiedWins = [];
+var tiedWinnersTemp = [];
+
+//Initial Game State
+var gameState = `summonTheGamemaster`;
+console.log(`gameState`, gameState);
```
DominiqueYeo, line 30: console.log should be removed from code  submissions

script.js
```diff
@@ -1,4 +1,343 @@
...
+var roundsPlayed = 0;
+
+// Main Function
 var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
+  input = input.toLowerCase();
```
DominiqueYeo, line 35: nice input formatting!

script.js
```diff
@@ -1,4 +1,343 @@
...
+  }
+
+  //Game settings code begins here
+
+  if (
+    gameState == `summonTheGamemaster` &&
+    input != `open the chamber of secrets`
+  ) {
+    return `The Gamemaster is quite a particular character. Say the words he wants or he will not entertain you.`;
+  }
```
DominiqueYeo, line 48: HAHA such a character this gamemaster. But good use of if and early return to control the flow of your code

script.js
```diff
@@ -1,4 +1,343 @@
...
+    gameState == `gameSettingsNumPlayers` &&
+    (input == 2 ||
+      input == `two` ||
+      input == 3 ||
+      input == `three` ||
+      input == 4 ||
+      input == `four` ||
+      input == 5 ||
+      input == `five`)
+  ) {
```
DominiqueYeo, line 98: you can look into using some array methods to help you reduce this .... in particular the .includes() method [link here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes)

DominiqueYeo: alternatively ... you can just reject the strings and only take numbers to make your life easier lol

script.js
```diff
@@ -1,4 +1,343 @@
...
+        ) {
+          playersDiceRolls = [];
+          for (
+            var diceRollCount = 0;
+            diceRollCount < numberOfDice;
+            diceRollCount += 1
+          ) {
+            var diceRoll = function () {
+              return Math.ceil(Math.random() * 6);
+            };
```
DominiqueYeo, line 230: i like that you shortened your diceRoll to something so short! but maybe move this out of the main function (functions declared before main )

script.js
```diff
@@ -1,4 +1,343 @@
...
+        return `At the end of ${roundsPlayed} exciting rounds, we finally have a winner!<br>
+<br>
+<b>Congratulations ${winner}</b>, luck was in your favour tonight and you won a total of ${maxArrayValue} rounds!<br>
+<br>
+If you wish to continue playing, just click the Start Over button above.<br>
+<br>
+If you've had enough, then I wish you farewell and hope we meet again soon.`;
+      }
+    }
+  }
```
DominiqueYeo, line 342: Since this two return messages are the same , maybe you can write a function for them ?   

` var winMessage = function (roundsPlayed,winner) { return message }`

then you can `return winMessage(roundsPlayed , winner)`

DominiqueYeo (commented):
Generally good commenting and naming for the variables. 

The next step would be to refactor your code so that your main function is shorter . 

a very ideal case is where you mostly only call functions in the main 

var main (input) {
doThis()
doThat()
thenDoThis()}


but dont force it and sacrifice readability

PR #361, opened 2022-03-07
Basic 13-2 Beat-that - James

rocketotter:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

4hrs

Comfort Level (1-5):
4
Completeness Level (1-5):
3
o,o" Can be better
Class implementation

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,6 @@
+// Testing 123!
+
 var main = function (input) {
-  var myOutputValue = 'hello world';
+  var myOutputValue = "hello world";
   return myOutputValue;
 };
--- a/script.js
+++ b/script.js
@@ -1,6 +1,259 @@
-// Testing 123!
+// /*
+// Included features:
+// - Last roll
+// - Winning %
+// - Most rolled
+// - Guessing +- 1
+// */
+
+/* 
+Fn:
+input: 
+- Number of users
+
+Each submit roll 2 dices
+track score
+*/
+var totalGames = 0;
+
+class player {
+  constructor(playername, dicenumber, score) {
+    this.playername = playername;
+    this.dicenumber = dicenumber;
+    this.combinedDiceNumber = this.combinedDiceNumber;
+    this.score = score;
+  }
+}
+
+function checkDiceForWinner() {
+  bblSortDice(plist);
+  plist[0].score += 1;
+}
+
+// BubbleSort directly to object.score highest to lowest (tbd)
+function bblSort(arr) {
+  for (var i = 0; i < arr.length; i++) {
+    // Last i elements are already in place
+    for (var j = 0; j < arr.length - i - 1; j++) {
+      // Checking if the item at present iteration
+      // is greater than the next iteration
+      if (arr[j].score > arr[j + 1].score) {
+        // If the condition is true then swap them
+        var temp = arr[j];
+        arr[j] = arr[j + 1];
+        arr[j + 1] = temp;
+      }
+    }
+  }
+}
+
+// BubbleSort directly to object.score highest to lowest (tbd)
+function bblSortDice(arr) {
+  for (var i = 0; i < arr.length; i++) {
+    // Last i elements are already in place
+    for (var j = 0; j < arr.length - i - 1; j++) {
+      // Checking if the item at present iteration
+      // is greater than the next iteration
+      if (arr[j].combinedDiceNumber > arr[j + 1].combinedDiceNumber) {
+        // If the condition is true then swap them
+        var temp = arr[j];
+        arr[j] = arr[j + 1];
+        arr[j + 1] = temp;
+      }
+    }
+  }
+}
+
+// Sort lowest to highest
+function bblSortNumberLow(arr) {
+  for (var i = 0; i < arr.length; i++) {
+    // Last i elements are already in place
+    for (var j = 0; j < arr.length - i - 1; j++) {
+      // Checking if the item at present iteration
+      // is greater than the next iteration
+      if (arr[j] > arr[j + 1]) {
+        // If the condition is true then swap them
+        var temp = arr[j];
+        arr[j] = arr[j + 1];
+        arr[j + 1] = temp;
+      }
+    }
+  }
+  // Print the sorted array
+  return arr;
+}
+
+// Sort array highest to lowest
+function bblSortNumberHigh(array) {
+  for (i = 0; i < array.length - 1; i++) {
+    for (j = 0; j < array.length - i - 1; j++) {
+      if (array[j] < array[j + 1]) {
+        var temp = array[j];
+        array[j] = array[j + 1];
+        array[j + 1] = temp;
+      }
+    }
+  }
+  return array;
+}
+
+function getCombinedDiceNumber(dice_arr, getHighLowOption) {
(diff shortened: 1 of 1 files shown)


script.js
```diff
@@ -1,4 +1,259 @@
...
+var totalGames = 0;
+
+class player {
+  constructor(playername, dicenumber, score) {
+    this.playername = playername;
+    this.dicenumber = dicenumber;
+    this.combinedDiceNumber = this.combinedDiceNumber;
+    this.score = score;
+  }
+}
```
lim-jiahao, line 26: nice use of classes to create player objects! good job going beyond what was taught

script.js
```diff
@@ -1,4 +1,259 @@
...
+      // is greater than the next iteration
+      if (arr[j].score > arr[j + 1].score) {
+        // If the condition is true then swap them
+        var temp = arr[j];
+        arr[j] = arr[j + 1];
+        arr[j + 1] = temp;
+      }
+    }
+  }
+}
```
lim-jiahao, line 48: great work implementing bubble sort but this, along with your other 3 sorting functions below, feel very repetitive

to do simple sorting like in the case here, you can leverage on javascript's own array.sort() method

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

lim-jiahao: for numbers, you would have to supply a comparison function as an argument into sort

lets say i have an array `var arr = [20, 10, 30, 40]`

```
arr.sort(function(a, b) {
  return a - b;
});
console.log(arr); // [10, 20, 30, 40]
```
basically, the comparison function takes in two arguments (which we call a and b) then it applies this sorting function to every pair of elements in the array. if the comparison function returns < 0, then .sort will place a before b, and b before a if the comparison function returns > 0

that's why when we return `a - b`, it has the effect of sorting it in ascending order. lets just take the first two elements as example, a will be 20 and b will be 10, so `a - b` will be 10 > 0, so 10 will be placed before 20

to sort in descending order, we just need to return `b - a`

lim-jiahao: lmk if you got any qns on this! i know it can be a complicated concept to grasp at first

script.js
```diff
@@ -1,4 +1,259 @@
...
+      // is greater than the next iteration
+      if (arr[j].combinedDiceNumber > arr[j + 1].combinedDiceNumber) {
+        // If the condition is true then swap them
+        var temp = arr[j];
+        arr[j] = arr[j + 1];
+        arr[j + 1] = temp;
+      }
+    }
+  }
+}
```
lim-jiahao, line 65: see comments above

script.js
```diff
@@ -1,4 +1,259 @@
...
+        // If the condition is true then swap them
+        var temp = arr[j];
+        arr[j] = arr[j + 1];
+        arr[j + 1] = temp;
+      }
+    }
+  }
+  // Print the sorted array
+  return arr;
+}
```
lim-jiahao, line 84: see comments above

script.js
```diff
@@ -1,4 +1,259 @@
...
+    for (j = 0; j < array.length - i - 1; j++) {
+      if (array[j] < array[j + 1]) {
+        var temp = array[j];
+        array[j] = array[j + 1];
+        array[j + 1] = temp;
+      }
+    }
+  }
+  return array;
+}
```
lim-jiahao, line 98: see comments above

script.js
```diff
@@ -1,4 +1,259 @@
...
+  for (let i = 0; i < numOfPlayers; i++) {
+    //    plist.push(new player("p" + i, bblSortNumber(roll_n_Dices(5)), rollDice()));   // For testing purpose, roll random 5 dices
+    plist.push(new player("p" + i, [], 0, 0));
+  }
+}
+
+function printScoreBoard(playerlists, mode) {
+  var output = "The score board now!<br>";
+  bblSort(plist); //sort the player list according to score lowest first
+  console.log(playerlists);
```
lim-jiahao, line 144: good use of console.log statements for debugging but they are usually removed before shipping to production in a professional setting (the equivalent of submitting your project for this course)

applies for your other console.log statements too

script.js
```diff
@@ -1,4 +1,259 @@
...
+      console.log(i);
+      output =
+        output +
+        "Playername: " +
+        playerlists[i].playername +
+        " " +
+        "Score is: " +
+        playerlists[i].score +
+        "<br>";
+    }
```
lim-jiahao, line 170: alternatively why dont we sort our player list in ascending or descending score order based on whether mode is reverse or not, then we only need one loop here where we are always printing from i = 0 to playerlists.length - 1

script.js
```diff
@@ -1,4 +1,259 @@
...
+      inputGameMode = false;
+      console.log("inputGameMode " + gameMode);
+      myOutputValue = `Please enter your if you wish to autocombine dicenumber to the highest or lowest: Input 'high or 'low'`;
+    } else {
+      console.log("inputGameMode " + gameMode);
+      autoCombineOption = input;
+      startOfGame = false;
+      myOutputValue = `Game start.<br> Number of players = ${numOfPlayers}. Number of dices = ${numOfDices}. <br> Game Mode is ${gameMode}Click submit to start!`;
+      playersCreation(numOfPlayers); //All players data created will be sorted in plist array
+    }
```
lim-jiahao, line 230: this is good but do play around with html and DOM manipulation in the future so that these modes can be entered using alternative means rather than free text input which would require input validation 

one way would be using additional buttons and input types

script.js
```diff
@@ -1,4 +1,259 @@
...
+    if (gameMode == "normal") {
+      console.log("normal mode");
+
+      // Each player roll a dice, den combine the dice number and update each player's details
+      for (let i = 0; i < plist.length; i++) {
+        plist[i].dicenumber.push(rollDice());
+        plist[i].combinedDiceNumber = getCombinedDiceNumber(
+          plist[i].dicenumber,
+          autoCombineOption
+        );
```
lim-jiahao, line 243: shouldnt this be called outside the loop, after it finishes? only after we get all dice rolls then we get the combined dice number

script.js
```diff
@@ -1,4 +1,259 @@
...
+      }
+      // roll dice, add dice to player profile, calculate score based on gameMode
+    }
+    if (gameMode == "reverse") {
+      console.log("reverse mode");
+    }
+
+    if (input == "reset") {
+      console.log("reset mode");
+    }
```
lim-jiahao, line 253: wait whats this lol

lim-jiahao:
hi James,

overall very good work! your grasp of the fundamentals is solid as i pointed out for your SPS project and its great that you're going beyond what is taught to implement classes and bubble sort. 

my comment from previously will still apply:
try playing around with the HTML (and CSS if you have time) for your next project. Currently when I play your game, I am greeted with an empty output box and it is not immediately clear what I have to do. Always take user experience into account when you code. A good way to improve on this is to simply edit the HTML so that a message is in the output box immediately when the game is loaded, perhaps to welcome the user and prompt the user to enter their name or give instructions on the rules of the game, how the game is played

Also try the more comfortable portions too! Last project already so push yourself more, you can do it 👍

PR #370, opened 2022-06-14
Fujinami Ryoji_15-7_Beat That

ryojif2:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

5 hours

Not being able to output an intermediate message to let the player choose the order of their numbers.

Figuring out how to solve the above error.

Comfort Level (1-5):
4

Completeness Level (1-5):
4

Decent.

Being able to use loop and array to keep a scoreboard.

How to make the code less clumsy.

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,252 @@
+//GLOBAL VARIABLE
+//game mode. waiting for game to start as default.
+//game step. Continuation from basic game structure. Allows for division of functions within each game mode.
+var gameMode = "waiting for game to start";
+var gameStep = "player 1 roll dice";
+
+//store player dice number with array.
+var playerNumber = [];
+
+//store player combined number with array.
+var player1CombinedNumber = [];
+var player2CombinedNumber = [];
+
+//global round counter.
+var roundCounter = 1;
+
+//FUNCTION 1
+// create roll dice function.
+var rollDice = function () {
+  // Generate a decimal from 0 through 6, inclusive of 0 and exclusive of 6.
+  var randomDecimal = Math.random() * 6;
+
+  // Remove the decimal with the floor operation.
+  // This will be an integer from 0 to 5 inclusive.
+  var randomInteger = Math.floor(randomDecimal);
+
+  // Add 1 to get valid dice rolls of 1 through 6 inclusive.
+  var diceNumber = randomInteger + 1;
+
+  return diceNumber;
+};
+
+//FUNCTION 2 - rolls Dice 1 and 2
+// call roll dice 1 & dice 2
+var playerDice = function () {
+  var dice1Number = rollDice();
+  var dice2Number = rollDice();
+
+  // produce both numbers in an array to player.
+  playerNumber = [dice1Number, dice2Number];
+};
+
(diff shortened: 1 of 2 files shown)


script.js
```diff
@@ -1,4 +1,252 @@
...
+};
+
+//FUNCTION 3 - Sum up player 1's numbers every round and 2's numbers every round, and announce winner for current round
+var sumPlayerNumbers = function () {
+  var winnerResult = "";
+  //create loop to sum up all previous rounds' numbers saved in the CombinedNumber array of each player.
+  var index = 0;
+  var player1NumberSum = 0;
+  var player2NumberSum = 0;
+  while (index < roundCounter) {
```
hpgay, line 50: nice use of loop to sum up both player 1 and 2's numbers

script.js
```diff
@@ -1,4 +1,252 @@
...
+
+  // Add 1 to get valid dice rolls of 1 through 6 inclusive.
+  var diceNumber = randomInteger + 1;
+
+  return diceNumber;
+};
+
+//FUNCTION 2 - rolls Dice 1 and 2
+// call roll dice 1 & dice 2
+var playerDice = function () {
```
hpgay, line 35: name functions starting with a verb, this is a javascript convention so that it's easier to tell functions from variables. in this case maybe something like, rollPlayerDice / roll2Dice

script.js
```diff
@@ -1,4 +1,252 @@
...
+//FUNCTION 4 - NORMAL GAME MODE
+var playNormalGame = function () {
+  var myOutputValue = "";
+  // call function 2 for player 1.
+  // automatically throw dice and assign number for most optimal result.
+  // let player 1 know the result.
+  // change game mode to player 2.
+  if (gameStep == "player 1 roll dice") {
+    gameStep = "change to player 2";
+    myOutputValue = playerDice();
```
hpgay, line 88: `playerDice()` is a function that doesn't return any value right? It just resets the playerNumber array if I'm not mistaken. so there's no need to assign` playerDice()` to the variable `myOutputValue`, you can just execute the function as is, i.e. this line would just be `playerDice()` and the global variable `playerNumber` will be updated.

script.js
```diff
@@ -1,4 +1,252 @@
...
+  }
+  // call function 2 for player 2.
+  // automatically throw dice and assign number for most optimal result.
+  // let player 2 know the result.
+  // change game mode to player 2.
+  else if (gameStep == "player 2 roll dice") {
+    myOutputValue = playerDice();
+    if (playerNumber[0] >= playerNumber[1]) {
+      player2CombinedNumber.push(
+        Number(playerNumber[0] * 10 + playerNumber[1])
```
hpgay, line 123: just fyi, another way to combine the two numbers without adding them up is to concatenate them as strings and then changing that number string into a Number

hpgay (commented):
Hi Ryoji! Everything looks good, good logic flow with comments that makes it easy to follow, good refactoring into separate functions. I've left some comments on specific lines of code that you might want to take note of.

PR #385, opened 2022-06-16
replaced files index.html and script.js

Bekokid:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

Comfort Level (1-5):

Completeness Level (1-5):

### DIFF
--- a/script.js
+++ b/script.js
@@ -18,6 +18,7 @@ var main = function (input) {
   var rollDice1;
   var rollDice2;
 
+  // clickCount
   if (clickCount % 2 == 0) {
     rollDice1 = getRandomIndex();
     rollDice2 = getRandomIndex();
@@ -65,16 +66,23 @@ var main = function (input) {
 
 var main01 = function () {
   var outputValue = "";
+  var resultsReport = "";
+  var resultList = [];
 
   var maxNumber = Math.max(...playerResults);
 
   outputValue = `🏆  The winner is player ${
     playerGroupNumber[playerResults.indexOf(maxNumber)]
-  }`;
+  } <br><br>The numbers of each playes are shown as below:<br>`;
+
+  for (var i = 0; i < playerResults.length; i++) {
+    resultsReport = `The number of player ${playerGroupNumber[i]} is ${playerResults[i]}.<br>`;
+    resultList.push(resultsReport);
+  }
 
   console.log(maxNumber);
   console.log(outputValue);
   console.log(playerResults);
   console.log(playerResults.indexOf(maxNumber) + 1);
-  return outputValue;
+  return outputValue + resultList;
 };
--- a/script.js
+++ b/script.js
@@ -4,6 +4,12 @@ var tempNum = [];
 var playerNumber = 1;
 var playerGroupNumber = [];
 
+// clickCount is to control the mode:
+// When clickCount equal to even number, it's "Dice-roll mode":
+// Generate 2 random numbers and show to players
+// When clickCount equal to odd number, it's "Push mode":
+// Finalise the number according to the order player choose, and push the number into array "playerResults"
+
 var getRandomIndex = function () {
   // create a random number from zero to array length minus one.
   // this number is in the range from the first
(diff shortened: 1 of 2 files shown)


script.js
```diff
@@ -1,4 +1,106 @@
...
+
+      // the generated random numbers are temporarily stored in array "tempNum"
+      tempNum = [rollDice1, rollDice2];
+    }
+
+    // Push mode: clickCount equal to odd number
+  } else {
+    //When player choose "1":
+    if (input == 1) {
+      playerNember = clickCount;
```
hpgay: is this a typo? should it have been `playerNumber` instead of `playerNember`?

hpgay: why do you assign `clickCount` to `playerNember` here?

script.js
```diff
@@ -1,4 +1,106 @@
...
+      currentNumber = tempNum[0] * 10 + tempNum[1];
+      playerResults.push(currentNumber);
+      myOutputValue = `The number player ${playerNumber} choose is ${currentNumber}. <br><br>Next player please click "Submit" to roll the dice. `;
+      playerGroupNumber.push(playerNumber);
+      playerNumber++;
+
+      //When player choose "2":
+    } else if (input == 2) {
+      playerNember = clickCount;
```
hpgay: why do you assign `clickCount` to `playerNember` here?

script.js
```diff
@@ -1,4 +1,106 @@
...
+
 var main = function (input) {
-  var myOutputValue = 'hello world';
+  var myOutputValue = "";
+  var currentNumber;
+  var rollDice1;
+  var rollDice2;
+
+  // Dice-roll mode: clickCount equal to even number
+  if (clickCount % 2 == 0) {
```
hpgay, line 28: it might be easier to read if you just use `mode == 'dice-roll mode'` to show the different modes, because words are easier to understand than to use even/odd number, especially when the clickCount keeps increasing

script.js
```diff
@@ -1,4 +1,106 @@
...
+  console.log("Dice 2 number: ", rollDice2);
+  console.log("Click count: ", clickCount);
+  console.log(tempNum);
+  console.log(playerResults);
+  clickCount++;
   return myOutputValue;
 };
+
+// Another function for button "Results" which is to show the winner and list out all the numbers
+var main01 = function () {
```
hpgay, line 82: use a different name from `main01` because it doesn't say what the function does, maybe something like `calculateResults`

script.js
```diff
@@ -1,4 +1,106 @@
...
+      // the generated random numbers are temporarily stored in array "tempNum"
+      tempNum = [rollDice1, rollDice2];
+    }
+
+    // Push mode: clickCount equal to odd number
+  } else {
+    //When player choose "1":
+    if (input == 1) {
+      playerNember = clickCount;
+      currentNumber = tempNum[0] * 10 + tempNum[1];
```
hpgay, line 51: i think there's no need to have `tempNum`, you can just use the values of `rollDice1` and `rollDice2` to figure out what the `currentNumber` is right?

hpgay (commented):
Hi Sern, I appreciate how you try different ways to structure your code. The general flow of logic is good. I also appreciate how you added another button so that the code becomes much simpler. See the specific comments that I left in the code for how some things can be improved.

script.js
Bekokid: Yes, that's a typo. 
And also the variable "playerNember" is no longer in use. It should be removed actually.

Bekokid: Variable "playerNember" is no longer in use. It should be removed actually.

script.js
Bekokid: Variable "playerNember" is no longer in use. It should be removed actually.

script.js
Bekokid: Noted

script.js
Bekokid: Noted

script.js
Bekokid: Actually, I tried to use the values of  `rollDice1` and `rollDice2` directly but failed.
I think it is because those 2 random numbers are generated in the previous mode (when `clickCount` % 2 == 0).
Then they are not able to be accessed in the following "Push mode".
And that's why I set `tempNum` to carry the random numbers across the modes.

PR #439, opened 2022-09-19
17-2 Darrick Beat-that

DarrickT:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

8.
Many !!
debugging.
Comfort Level (1-5):
2
Completeness Level (1-5):

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,207 @@
+var GAME_MODE_CHOOSE_NUM_PLAYERS = "GAME_MODE_CHOOSE_NUM_PLAYERS";
+var GAME_MODE_CHOOSE_NUM_DICE = "GAME_MODE_CHOOSE_NUM_DICE";
+var GAME_MODE_DICE_ROLL = "GAME_MODE_DICE_ROLL";
+var GAME_MODE_CHOOSE_DICE_ORDER_AUTOMATICALLY =
+  "GAME_MODE_CHOOSE_DICE_ORDER_AUTOMATICALLY";
+
+var numDiceChosen = 0;
+
+var numPlayers = 0;
+
+var gameMode = GAME_MODE_CHOOSE_NUM_PLAYERS;
+
+var currPlayer = 0;
+
+var playerProfiles = [];
+
+var roundWinner = null;
+
+var getDiceRoll = function () {
+  return Math.ceil(Math.random() * 6);
+};
+
+var getDiceRolls = function () {
+  var newDiceRolls = [];
+  for (var i = 0; i < numDiceChosen; i += 1) {
+    newDiceRolls.push(getDiceRoll());
+  }
+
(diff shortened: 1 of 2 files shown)


script.js
```diff
@@ -1,4 +1,207 @@
...
+  for (var i = 0; i < numDiceChosen; i += 1) {
+    newDiceRolls.push(getDiceRoll());
+  }
+
+  playerProfiles[currPlayer].diceRolls = newDiceRolls;
+
+  return newDiceRolls;
+};
+
+var concatenate2Numbers = function (num1, num2) {
```
zephaniahong, line 34: This is a really good function! Well done(:

script.js
```diff
@@ -1,4 +1,207 @@
...
+  numDiceChosen = 0;
+};
+//
 var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
+  console.log(gameMode);
+  console.log(currPlayer);
+  if (gameMode == GAME_MODE_CHOOSE_NUM_PLAYERS) {
+    if (isNaN(input) == true) {
```
zephaniahong, line 117: you should never need to evaluate if something == true. In most, if not all cases, you can do without it. e.g. isNaN(input) will already resolve to a true or false.

script.js
```diff
@@ -1,4 +1,207 @@
...
+      return "Please enter a number between 2 players to 5 players";
+    }
+    if (Number(input) < 2) {
```
zephaniahong, line 120: You seem to be converting input to Number very often and then comparing it to another Number. You can consider creating a function for this. e.g.
```js
function isLessThan(input, num) {
// logic here
}
```

script.js
```diff
@@ -1,4 +1,207 @@
...
+
+    gameMode = GAME_MODE_CHOOSE_NUM_DICE;
+    return (
+      "There are " +
+      numPlayers +
+      " players in this game. Please enter how many dice you would like to play with."
+    );
+  }
+
+  if (gameMode == GAME_MODE_CHOOSE_NUM_DICE) {
```
zephaniahong, line 137: I noticed you're only using if statements in your main function. Try to use else if where appropriate as well. Having multiple ifs like this signals to the reader that more than 1 if could potentially happen. But in your case, that's not what is really happening

zephaniahong (commented):
Hey Darrick! Overall, a really good job with this project! Can see that you're very comfortable with creating helper functions and using them in a good way as well. One thing you could improve on would be to try to keep your main function cleaner. There is still a lot of logic in the main function that could be refactored into a helper function.

script.js
```diff
@@ -1,4 +1,207 @@
...
+    if (isNaN(input) == true) {
```
zephaniahong, line 138: A lot of the following logic can be refactored into a helper function

PR #454, opened 2023-01-05
Basics 12-3 Beat That Submission: Irvin

irvincodes:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

5

Comfort Level (1-5):
3.5
Completeness Level (1-5):
4
With the video, it helped and was a good guide to follow

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,93 @@
+// === REQUIREMENTS ===
+// 1. 2 players play and they take turns
+// 2. when a player clicks submit, the game rolls 2 dice and show the output of the dice rolls e.g 2 and 4
+// 3. that player picks the order of the dice they want. e.g if he wanted 42, specify 2nd dice goes first
+// 4. after both players rolled and chosen dice order, scores are compared and higher number wins
+
+// === problem breakdown and planning ===
+// ver 1. rolls 2 dice and returns the output for 1 player. that player chooses the dice order and gets the correct output value returned
+// ver 2. refactor code to include player 2's turn
+// ver 3. implement comparing dice scores and declare winner
+// ver 4. reset game so that players can play again without refreshing browser page
+
+// === code start ===
+// global variables
+var GAME_STATE_DICE_ROLL = "GAME_STATE_DICE_ROLL";
+var GAME_STATE_CHOOSE_DICE_ORDER = "GAME_STATE_CHOOSE_DICE_ORDER";
+var gameState = GAME_STATE_DICE_ROLL;
+
+var playerRolls = [];
+
+// helper function
+var rollDice = function () {
+  console.log("Control Flow: Start of rollDice()");
+  var randomDecimal = Math.random() * 6;
+  var randomInteger = Math.floor(randomDecimal) + 1;
+  console.log("rollDice output, randomInteger: " + randomInteger);
+  return randomInteger;
+};
+
+var rollDiceForPlayer = function () {
+  console.log("Control Flow: Start of rollDiceForPlayer()");
+  for (let i = 0; i < 2; i += 1) {
+    playerRolls.push(rollDice());
+  }
+  console.log("rollDiceForPlayer changes, playerRolls: " + playerRolls);
+  return (
+    "Welcome!<br><br>You rolled:<br>Dice 1: " +
+    playerRolls[0] +
+    " | Dice 2: " +
+    playerRolls[1] +
+    ".<br><br>Sweet! Now input either '1' or '2' to choose the corresponding dice to be used as the first digit of your final value."
+  );
+};
+
+var getPlayerScore = function (playerInput) {
+  // input validation
+  if (playerInput != 1 && playerInput != 2) {
+    console.log(
+      "Control Flow: Input validation, invalid input... INPUT NOT 1 OR 2"
+    );
+    return (
+      "Invalid input! Please only input '1' or '2' to choose which dice to use as the first digit.<br><br>Your dice rolls are:<br>Dice 1: " +
+      playerRolls[0] +
+      " | Dice 2: " +
+      playerRolls[1] +
+      "."
+    );
+  }
+  // input == 1
+  if (playerInput == 1) {
+    console.log("Control Flow: input == 1");
+    var playerScore = Number(String(playerRolls[0]) + String(playerRolls[1]));
+    return "Your chosen value is: '" + playerScore + "'.";
+  }
+  // input == 2
+  if (playerInput == 2) {
+    console.log("Control Flow: input == 2");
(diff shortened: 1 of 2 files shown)


script.js
```diff
@@ -1,4 +1,165 @@
...
+var GAME_STATE_COMPARE_SCORES = "GAME_STATE_COMPARE_SCORES";
+var gameState = GAME_STATE_DICE_ROLL;
+
+var currentPlayerRolls = [];
+
+var currentPlayer = 1;
+var allPlayersScore = [];
+
+// helper function
+var rollDice = function () {
```
zephaniahong, line 29: You can combine all the steps into 1
```js
return Math.floor(Math.random() * 6) + 1
```

script.js
```diff
@@ -1,4 +1,165 @@
...
+  console.log("Control Flow: Start of rollDice()");
+  var randomDecimal = Math.random() * 6;
+  var randomInteger = Math.floor(randomDecimal) + 1;
+  console.log("rollDice output, randomInteger: " + randomInteger);
+  return randomInteger;
+};
+
+var rollDiceForPlayer = function () {
+  console.log("Control Flow: Start of rollDiceForPlayer()");
+  for (let i = 0; i < 2; i += 1) {
```
zephaniahong, line 39: Extra: You could consider making a variable: numberOfPlayers, and then use this variable in your for loop instead of hardcoding 2 here. Reason for this is that in the future, if you would like to increase the number of players, you can simply change the value of the variable instead of finding all the instances where you hardcoded 2 and replacing it.

script.js
```diff
@@ -1,4 +1,165 @@
...
+      currentPlayerRolls[0] +
+      " | Dice 2: " +
+      currentPlayerRolls[1] +
+      "."
+    );
+  }
+  // input == 1
+  if (playerInput == 1) {
+    console.log("Control Flow: input == 1");
+    var playerScore = Number(
```
zephaniahong, line 73: This pattern of casting to String and then to Number is repeated several times. Would be good to have a helper function that can do all the necessary casting for you.

script.js
```diff
@@ -1,4 +1,165 @@
...
+};
+
+var comparePlayerScores = function () {
+  compareMessage =
+    "Player 1 Score: " +
+    allPlayersScore[0] +
+    "<br>Player 2 Score: " +
+    allPlayersScore[1];
+  // player 1 wins
+  if (allPlayersScore[0] > allPlayersScore[1]) {
```
zephaniahong, line 100: It would be more ideal to use if, else if and else here even though technically it doesn't make any difference.

script.js
```diff
@@ -1,4 +1,165 @@
...
+    compareMessage = compareMessage + "<br><br>Player 2 wins!";
+  }
+  // tie
+  if (allPlayersScore[0] == allPlayersScore[1]) {
+    compareMessage = compareMessage + "<br><br>It's a tie!";
+  }
+  return compareMessage;
+};
+
+var resetGame = function () {
```
zephaniahong, line 114: Nice helper function!

script.js
```diff
@@ -1,4 +1,165 @@
...
+  return "<br><br>Click submit to play again!";
+};
+
 var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
+  console.log("Check game state on submit click: " + gameState);
+  console.log("Checking currentPlayer on submit click: " + currentPlayer);
+  var outputMessage = "";
+  if (gameState == GAME_STATE_DICE_ROLL) {
```
zephaniahong, line 125: Similarly, for you game states, use if, else if and else statements

script.js
```diff
@@ -1,4 +1,165 @@
...
+    console.log("Control Flow: gameState == GAME_STATE_DICE_ROLL");
+
+    // display dice numbers rolled as output message
+    outputMessage = rollDiceForPlayer();
```
zephaniahong, line 129: One thing I noticed is that your helper functions tend to return strings for your output message. But if you take a look at your function naming, it does not seem to convey that meaning. 
As a reader of your code, I would expect rollDiceForPlayer to only roll the dice and maybe store the dice value in some array. I would not expect it to return me an entire string for the output message.
I'll talk more about this in class today

script.js
```diff
@@ -1,4 +1,165 @@
...
+    gameState = GAME_STATE_CHOOSE_DICE_ORDER;
+    return outputMessage;
+  }
+  if (gameState == GAME_STATE_CHOOSE_DICE_ORDER) {
+    console.log("Control Flow: gameState == GAME_STATE_CHOOSE_DICE_ORDER");
+
+    // call getPlayerScore function
+    outputMessage = getPlayerScore(input);
+
+    if (currentPlayer == 1) {
```
zephaniahong, line 141: use if and else

script.js
```diff
@@ -1,4 +1,165 @@
...
+        outputMessage +
+        "<br><br>Click submit to calculate scores and see who won!"
+      );
+    }
+  }
+  if (gameState == GAME_STATE_COMPARE_SCORES) {
+    console.log("Control Flow: gameState == GAME_STATE_COMPARE_SCORES");
+
+    outputMessage = comparePlayerScores() + resetGame();
+    resetGame();
```
zephaniahong, line 162: this is redundant as you already reset the game on line 161

script.js
```diff
@@ -1,4 +1,165 @@
...
+      return (
+        outputMessage +
+        "<br><br>Click submit to calculate scores and see who won!"
+      );
+    }
+  }
+  if (gameState == GAME_STATE_COMPARE_SCORES) {
+    console.log("Control Flow: gameState == GAME_STATE_COMPARE_SCORES");
+
+    outputMessage = comparePlayerScores() + resetGame();
```
zephaniahong, line 161: This is not the ideal way to use functions. I'll talk more about this in class

script.js
```diff
@@ -1,4 +1,165 @@
...
+// ver 1. rolls 2 dice and returns the output for 1 player. that player chooses the dice order and gets the correct output value returned
+// ver 2. refactor code to include player 2's turn
+//      - global variables for currentPlayer; allPlayersScore
+//      - refactor outputMessages to interact with each player 1 and 2
+//      - write logic for player 1 to go first then player 2, then proceed to compare score
+// ver 3. implement comparing dice scores and declare winner
+// ver 4. reset game so that players can play again without refreshing browser page
+
+// === code start ===
+// global variables
```
zephaniahong, line 17: Good organisation of global variables

zephaniahong (commented):
Hey Irvin! Overall, a really good job with this project! It's clear that you have a good understanding of control flow and the use of helper functions. The use of arrays are good as well!

PR #458, opened 2023-01-10
Wei Wei's Beat That Submission

0x1p:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

How many hours did you spend on this assignment? a few hours. mostly thinking of the logic.

Please fill in one error and/or error message you received while working on this assignment. usually i get caught on the = vs. == error

What part of the assignment did you spend the most time on? sorting and translating strings/arrays to number

Comfort Level (1-5): 4

Completeness Level (1-5): 4

What did you think of this deliverable? good progress

Is there anything in this code that you feel pleased about? my convertArrayToNumber()

What's one aspect of your code you would like specific, elaborate feedback on? my while loop sorting function below - how to fix it?

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,230 @@
-var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
+//Cloned from day9 more dice more players
+
+//GLOBAL VARIABLES
+//Game states
+var START = "START";
+var SET_PLAYERS = "SET_PLAYERS";
+var SET_ROLLS = "SET_ROLLS";
+var DICE_ROLL = "DICE_ROLL";
+var SCOREBOARD = "SCOREBOARD";
+
+//Selected number of players and dice to play with
+var playersList = [];
+var chosenDiceRolls = 0;
+
+//Set initial state
+var gameState = START;
+var playerInPlay;
+var roundWinner;
+var overallWinner;
+
+//HELPER FUNCTIONS
+//create an array of objects - # of objects = # of players
+var createPlayersList = function (input) {
+  var currentPlayer = 1;
+
+  while (currentPlayer <= input) {
+    var player = {
+      name: `Player ${currentPlayer}`,
+      hand: [],
+      tally: 0,
+      score: 0,
+    }
+    playersList.push(player);
+    currentPlayer +=1;
+  }
+  return playersList;
 };
+
+
+//roll dice (1 to 6) for as many times as chosenDiceRolls indicates
+var rollDiceForPlayer = function () {
+  var counter = 0;
+
+  while (counter < chosenDiceRolls) {
+    playerInPlay.hand.push(Math.floor(Math.random() * 6) + 1);
+    counter += 1;
+  }
+  return playerInPlay.hand;
+};
+
+//sort largest to smallest number
+var sortBigToSmall = function (input) {
+
+  for (i = 0; i < input.length; i += 1) {
+    for (j = 0; j < (input.length - i - 1); j ++) {
+      if (input[j] < input[j+1]) {
+        var temp = input[j];
+        input[j] = input[j+1];
+        input[j+1] = temp;
+      } 
+    }
+      console.log(input)
+      console.log(j)
+      console.log(i)
+  }
+  return null;
+}
+
+//Converts array of numbers to a number
+var convertArrayToNumber = function (input) {
+  var runDown = chosenDiceRolls - 1;
+  var counter = 0;
+
+  while (runDown >= 0) {
+    var number = input[counter] * Math.pow(10, runDown);
+    console.log(number);
+    playerInPlay.tally += number;
+    console.log(playerInPlay.tally);
+    runDown -= 1;
+    counter += 1;
+  }
+  return playerInPlay.tally;
+};
+
+//Compares the score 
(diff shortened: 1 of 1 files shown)


script.js
```diff
@@ -1,4 +1,230 @@
...
+        var temp = playerInPlay.hand[j];
+        playerInPlay.hand[j] = playerInPlay.hand[j+1];
+        playerInPlay.hand[j+1] = temp;
+      }
+      j +=1;
+    }
+    i +=1;
+  }
+  return null;
+}
```
floatingtales, line 224: I think the reason why this sort doesn't work is that there is no `playerInPlay.hand` playerInPlay is only an array of numbers.

script.js
```diff
@@ -1,4 +1,230 @@
+//Cloned from day9 more dice more players
+
+//GLOBAL VARIABLES
+//Game states
+var START = "START";
+var SET_PLAYERS = "SET_PLAYERS";
+var SET_ROLLS = "SET_ROLLS";
+var DICE_ROLL = "DICE_ROLL";
+var SCOREBOARD = "SCOREBOARD";
```
floatingtales, line 9: Great usage of constants! It's also nice of you to use SCREAMING_SNAKE_CASE as it is the standard case to use for these kinds of variables

script.js
```diff
@@ -1,4 +1,230 @@
...
+      name: `Player ${currentPlayer}`,
+      hand: [],
+      tally: 0,
+      score: 0,
+    }
+    playersList.push(player);
+    currentPlayer +=1;
+  }
+  return playersList;
+};
```
floatingtales, line 37: Great way of:
- having objects as players so you can have the tally and score!
- having a loop to create players!

This is one of a few times that I've seen objects here and I like it!

script.js
```diff
@@ -1,4 +1,230 @@
...
+var sortBigToSmall = function (input) {
+
+  for (i = 0; i < input.length; i += 1) {
+    for (j = 0; j < (input.length - i - 1); j ++) {
+      if (input[j] < input[j+1]) {
+        var temp = input[j];
+        input[j] = input[j+1];
+        input[j+1] = temp;
+      } 
+    }
```
floatingtales, line 61: second time I've seen bubble sort used in all the time I've been teaching basics! this is great!

script.js
```diff
@@ -1,4 +1,230 @@
...
+
+  while (runDown >= 0) {
+    var number = input[counter] * Math.pow(10, runDown);
+    console.log(number);
+    playerInPlay.tally += number;
+    console.log(playerInPlay.tally);
+    runDown -= 1;
+    counter += 1;
+  }
+  return playerInPlay.tally;
```
floatingtales, line 82: This is a very novel way of adding in the values of the sorted number as an array! Great job!

script.js
```diff
@@ -1,4 +1,230 @@
...
+      roundWinner = roundWinner;
+    } else {
+      roundWinner = playersList[i+1];
+    }
+    i +=1;
+  }
+
+  roundWinner.score +=1;
+  console.log (playersList.score);
+  return roundWinner;
```
floatingtales, line 101: There are more advanced ways of finding the maximum of the scores, and finding the index of that particular winner, with the built-in javascript stuff like `Math.max` and `Array.indexOf`. It could make your code more compact! But this is a perfectly adequate way of finding who has the maximum tally!

script.js
```diff
@@ -1,4 +1,230 @@
...
+
+  if (gameState == DICE_ROLL) {
+    var playerCounter = 0;
+    var outputStatement = 'Let the games begin (again).<br>';
+
+    while (playerCounter < playersList.length) {
+      playerInPlay = playersList[playerCounter];
+
+      rollDiceForPlayer();
+      var copiedHand = [...playerInPlay.hand];
```
floatingtales, line 167: Nice to see that you're copying an array with this method! There are other ways to do it as well, like with `Array.slice`, but this is the method I prefer the most!

floatingtales (commented):
Great job on completing this assignment, Wei! This has got to be one of the most complete beat that project that I've seen! You're well on your way of doing great here! A couple notes:

- It's best to remove console logs on final submission, it's usually used for debugging purposes. And is generally left out of the final product.
- It's a matter of personal preference and standards, but on a lot of cases, for loops are the more preferred types of loops as opposed to while loops. I noticed that you're doing a lot of while loops as opposed to for loops. Which is fine, but you might encounter more for loops instead of while loops on production-level codes. The easiest way to think about when to use one loop and not the other is:
1. Does it have a counter? 
2. Does that counter go up or down by 1 each loop no matter what?
If the answer to both of them are yes, a for loop is more generally preferred. Otherwise, while loop is more generally preferred.

You're doing really great so far! Can't wait to see what you are going to do on your blackjack project!

script.js
```diff
@@ -1,4 +1,230 @@
...
+
+  while (i < playersList.length - 1) {
+    if (overallWinner.score > playersList[i+1].score) {
+      overallWinner = overallWinner;
+    } else {
+      overallWinner = playersList[i+1];
+    }
+    i +=1;
+  }
+
```
floatingtales, line 117: Same comment on the tally! This is a good way of finding the maximum score

PR #461, opened 2023-02-16
Lee Weirong Fundamentals 20-2 [Beat That!]

LAMEfreak:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

**5+ hours on Base version + Comfortable + Score / LeaderBoard versions** 

**I forgot to consider the input value as a string and not a number - was stuck trying to find the bug for a while.** 

**Refactoring the code to make the main function as succinct / readable as possible and researching new array methods.**

Comfort Level (1-5): 
**4**

Completeness Level (1-5): 
**5**

**Very do-able, just need to be patient and work at it step by step. Though not sure if there are more efficient ways or array methods that I could've used.**

**(1) Using the .reduce array method for the leaderboard portion (2) updating the DOM for each gameMode so that the button's text is updated according to the context**

**N/A**


script-1.js
```diff
@@ -0,0 +1,165 @@
+// Global variables to track game mode and players' dice rolls
+let gameMode = "roll1"; // other states: 'order1', 'roll2', 'order2', 'result', 'end'
```
zephaniahong, line 2: As a general point for all your global variables, they should all be `const` unless you specifically reassign them to another value later on in your code

script-1.js
```diff
@@ -0,0 +1,165 @@
...
+
+let p1DiceRolls = [];
+let p1Result;
+let p1Score = [];
+
+let p2DiceRolls = [];
+let p2Result;
+let p2Score = [];
+
+let btn = document.getElementById("submit-button");
```
zephaniahong, line 13: This should be const

script-1.js
```diff
@@ -0,0 +1,165 @@
...
+}
+
+function isValid(number) {
+  return number === "1" || number === "2";
+}
+
+function runOrderLogic(order) {
+  if (gameMode === "order1") {
+    if (order === "1") {
+      p1Result = Number(p1DiceRolls[0] + p1DiceRolls[1]);
```
zephaniahong, line 101: This line is being repeated 4 times which is a hint that maybe you could create a function for it

script2.js
```diff
@@ -0,0 +1,146 @@
...
+}
+
+function isValid(num) {
+  return num >= 1
+}
+
+function runRollLogic() {
+  for (let i = 0; i < diceRolls; i++) {
+    if (gameMode === "roll1") {
+      p1DiceRolls.push(Number(getDiceRoll()));
```
zephaniahong, line 124: these 2 lines are being repeated as well. Could you think of how you could create a generic function that could handle this?

script2.js
```diff
@@ -0,0 +1,146 @@
...
+  }
+}
+
+function compareNums (a, b) {
+  return b - a
+}
+
+function generateOptimalNum() {
+  if (gameMode === 'roll1') {
+    p1SortedDiceRolls = p1SortedDiceRolls.sort(compareNums);
```
zephaniahong, line 139: Same here. Try creating a function to handle this

script2.js
```diff
@@ -0,0 +1,146 @@
...
+  }
+}
+
+function updateScore() {
+  p1Score.push(p1Result);
+  p2Score.push(p2Result);
+}
+
+function showLeaderBoard() {
+  let p1Total = p1Score.reduce(
```
zephaniahong, line 93: You could create a function called sum so that you don't have to write the reduce twice since reduce is not as readable as a function called sum

script2.js
```diff
@@ -0,0 +1,146 @@
...
+    (accumulator, currentValue) => accumulator + currentValue
+  );
+  let p2Total = p2Score.reduce(
+    (accumulator, currentValue) => accumulator + currentValue
+  );
+  if (p1Total > p2Total) {
+    return `Leaderboard: <br><br> Player 1's cumulative score: ${p1Total}<br>Player 2's cumulative score: ${p2Total}<br><br><strong>Player 1</strong> is leading!<br><br>Enter the no. of dice to roll to play again.`;
```
zephaniahong, line 100: There's a lot of repeated text here. Try creating a function that takes in the scores and player id

zephaniahong (commented):
Hi Wei Rong! Great job on finishing your second project! It's a really big milestone so do give yourself credit for it. Some areas of improvement would be to use more wishful thinking. I can see that you are still writing majority of your logic in the main function, resulting in it being very long. Ideally, the main function should be as short as possible. 
Also, most of your functions dont take in any parameters. This usually means that those functions are very limited in their use case. There are quite a number of lines being repeated. Try to think how you could create a function for those repetitions. Other than that, you're on the right track!

PR #465, opened 2023-02-22
MarkChan_20-2_BeatThat

markcwy-ra:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

6

While doing the "More Comfortable" version of the assignment, I had to sort an array using .sort() (which I assigned to a new variable) to get the largest value, but when I tried searching for it in the original array, it kept returning the winner as Player 1.

I spent the most time on the "More Comfortable" section trying to sort the array from largest value to smallest. Didn't know there was a .sort() function so I tried using loops within loops to compare values and update index numbers in an effort to rearrange them. Eventually gave up and looked at the solution and realised there was a sort function 😅.

Comfort Level (1-5):
4

Completeness Level (1-5):
4

Felt like the more comfortable versions made use of loops more, which was more fun! 

I finally managed to make my main functions cleaner! 

How are my main functions looking? Haha. Tried to use more wishful thinking, but realised I still did do a little more back and forth to get it to be the most "wishful".

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,131 @@
-var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
-};
+let turn, mode;
+let player1Rolls, player2Rolls;
+let player1Number, player2Number;
+let p1score = 0;
+let p2score = 0;
+
+resetGame();
+
+function main(input) {
+  var outputValue;
+  if (mode === "roll") {
+    outputValue = getPlayerRolls(turn);
+  } else if (mode === "arrange") {
+    if (!isValid(input)) {
+      return invalidStatement();
+    } else {
+      outputValue = arrangedPlayerNumber(turn, input);
+      changeTurn();
+    }
+  } else if (mode === "compare") {
+    outputValue = winStatement(player1Number, player2Number) + scoreChart();
+    resetGame();
+  }
+  return outputValue;
+}
+
+function getPlayerRolls(turn) {
+  var player = generateDiceRolls(2);
+  changeMode("arrange");
+  if (turn === "Player 1") {
+    player1Rolls = player;
+  } else if (turn === "Player 2") {
+    player2Rolls = player;
+  }
+  return rollStatement(turn, player);
+}
+
+function arrangedPlayerNumber(turn, input) {
+  if (turn === "Player 1") {
+    player1Number = arrangeNumbers(player1Rolls, input);
+    changeMode("roll");
+    return arrangedStatement(turn, player1Number);
+  } else if (turn === "Player 2") {
+    player2Number = arrangeNumbers(player2Rolls, input);
+    changeMode("compare");
+    return arrangedStatement(turn, player2Number);
+  }
+}
+
+function resetGame() {
+  turn = "Player 1";
+  mode = "roll";
+  player1Rolls = [];
+  player2Rolls = [];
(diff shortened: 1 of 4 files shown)


script.js
```diff
@@ -1,4 +1,131 @@
...
-  var myOutputValue = 'hello world';
-  return myOutputValue;
-};
+let turn, mode;
+let player1Rolls, player2Rolls;
+let player1Number, player2Number;
+let p1score = 0;
+let p2score = 0;
+
+resetGame();
```
zephaniahong, line 7: Rather than calling this function, all the initial values of your globals should already be set

script.js
```diff
@@ -1,4 +1,131 @@
...
+
+function main(input) {
+  var outputValue;
+  if (mode === "roll") {
+    outputValue = getPlayerRolls(turn);
```
zephaniahong, line 12: Similar to what we discussed the last time for SPS, our functions should achieve the smallest task possible. In this case, your function is handling the player rolls as well as generating the message. This makes the function's reusability limited.

script.js
```diff
@@ -1,4 +1,131 @@
...
+  } else if (mode === "arrange") {
+    if (!isValid(input)) {
+      return invalidStatement();
+    } else {
+      outputValue = arrangedPlayerNumber(turn, input);
```
zephaniahong, line 17: Same here. The function is handling 2 tasks

script.js
```diff
@@ -1,4 +1,131 @@
...
+      changeTurn();
+    }
+  } else if (mode === "compare") {
+    outputValue = winStatement(player1Number, player2Number) + scoreChart();
```
zephaniahong, line 21: This is good!

script.js
```diff
@@ -1,4 +1,131 @@
...
+    rolls.push(diceRoll(6));
+  }
+  return rolls;
+}
+
+function diceRoll(number) {
+  return Math.ceil(Math.random() * number);
+}
+
+function arrangeNumbers(rolls, selection) {
```
zephaniahong, line 121: This is good as well!

script2.js
```diff
@@ -0,0 +1,125 @@
+let noOfPlayers, noOfDice;
+let diceRolls = [];
+let optimumNumbers = [];
+let mode = "playerCount";
+changeInstructions(mode);
```
zephaniahong, line 5: Similar to your earlier version, rather than calling this function here, you should have set the default look you needed in the html

script2.js
```diff
@@ -0,0 +1,125 @@
...
+    mode = "diceCount";
+  } else if (mode === "diceCount") {
+    if (!isValid(input)) {
+      return `Please enter a valid number!`;
+    }
+    noOfDice = input;
+    outputValue = `Chosen number of dice: ${noOfDice}. Click "Submit" to roll!`;
+    mode = "rollDice";
+  } else if (mode === "rollDice") {
+    diceRolls = generateDiceRolls(noOfPlayers, noOfDice);
```
zephaniahong, line 24: These are much better. Each function just handles a specific task

zephaniahong (commented):
Hi Mark! Congrats on completing your second project! Overall, its very well done and I don't have much comments for you. Just remember that your functions should only accomplish a single task. Other than that, you did a very good job!

PR #470, opened 2023-02-23
20-1 Diron 

Rjohnahld:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

1hr

finding out the player that wins

the switch case

Comfort Level (1-5):
3
Completeness Level (1-5):
3

my switch

### DIFF
--- a/script.js
+++ b/script.js
@@ -40,11 +40,11 @@ var main = function (input) {
           String(Math.max(...playerNumbers))
         );
         console.log(maxIndex);
-        var playingPlayer = [3, 2, 1];
+        var playingPlayer = [1, 2, 3];
         console.log(playingPlayer);
         console.log(playerNumbers);
         var winningPlayer = playingPlayer[maxIndex];
-        outputValue = `The winning number is ${maxValue} by ${winningPlayer}`;
+        outputValue = `The winning number is ${maxValue} by player ${winningPlayer}`;
         return outputValue;
     }
   }
--- a/script.js
+++ b/script.js
@@ -36,7 +36,9 @@ var main = function (input) {
         return outputValue;
       case 0:
         var maxValue = Math.max(...playerNumbers);
-        var maxIndex = playerNumbers.indexOf(Math.max(...playerNumbers));
+        var maxIndex = playerNumbers.indexOf(
+          String(Math.max(...playerNumbers))
+        );
         console.log(maxIndex);
         var playingPlayer = [3, 2, 1];
         console.log(playingPlayer);
--- a/script.js
+++ b/script.js
@@ -1,4 +1,73 @@
+var numOfPlayers = 3;
+var player = [3, 2, 1];
+var playerInputMessage = `please input "1" or "2" for the order of the dice`;
+var playerNumbers = [];
+var outputValue = "";
+var gameMode = "starting";
+var rng = function () {
+  var randomDecimal = Math.random() * 6;
+  var randomInteger = Math.floor(randomDecimal) + 1;
+  return randomInteger;
+};
+var diceRoll1 = rng();
+var diceRoll2 = rng();
+
 var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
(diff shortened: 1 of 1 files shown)


script.js
```diff
@@ -1,4 +1,75 @@
...
+  return randomInteger;
+};
+var diceRoll1 = rng();
+var diceRoll2 = rng();
+
 var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
+  if (gameMode == "starting") {
+    switch (numOfPlayers) {
```
floatingtales, line 17: Hey! You've learnt a little bit on switch cases! This is a great way of implementing a switch case, buddy!

script.js
```diff
@@ -1,4 +1,75 @@
...
+var playerNumbers = [];
+var outputValue = "";
+var gameMode = "starting";
+var rng = function () {
+  var randomDecimal = Math.random() * 6;
+  var randomInteger = Math.floor(randomDecimal) + 1;
+  return randomInteger;
+};
+var diceRoll1 = rng();
+var diceRoll2 = rng();
```
floatingtales, line 13: I don't think you need to assign this value in this variable here, you can just leave it empty first, and when you reached the main function, where it is actually used, only then you call the rng function.

script.js
```diff
@@ -1,4 +1,75 @@
...
+        diceRoll2 = rng();
+        numOfPlayers = numOfPlayers - 1;
+        outputValue = `player 3 dice roll one is ${diceRoll1} and dice roll two is ${diceRoll2}, ${playerInputMessage} `;
+        gameMode = "input 1 or 2";
+        return outputValue;
+      case 0:
+        var maxValue = Math.max(...playerNumbers);
+        var maxIndex = playerNumbers.indexOf(
+          String(Math.max(...playerNumbers))
+        );
```
floatingtales, line 41: Hey! I like how you're using `Math.max()` and `indexOf()`, But I think line 39 can be simplified to:
`var maxIndex = playerNumber.indexOf(String(maxValue))`
As you've already got the value!

script.js
```diff
@@ -1,4 +1,75 @@
...
+        console.log(playerNumbers);
+        var winningPlayer = playingPlayer[maxIndex];
+        outputValue = `The winning number is ${maxValue} by player ${winningPlayer}`;
+        return outputValue;
+    }
+  }
+  if (gameMode == "input 1 or 2") {
+    if (input == 1) {
+      var numberArray = player;
+      var players = Math.min(...numberArray);
```
floatingtales, line 54: I don't really think it's completely necessary to use a `Math.min()` in this instance, the player that is the lowest is the last index of the array, so you can just do something like `player[player.length-1]`, or if you want to be fancy and use a built in array function, `player.at(-1)`.

The at function functions similarly to the square bracket notation, but the key difference is that, if you put in a negative number there, it counts backwards from the end of the array, so:
`player[player.length-2]` is the same as `player.at(-2)`

floatingtales (commented):
Great work on this project Diron! You've got a novel way of solving this project! It's unlike a lot of the projects that I've seen. There are some pointers tho:

- You don't really need to have most of your variables defined at a global scope, if possible, minimize this. Global variables are easily manipulated in the console, whereas functional/code block scoped variables are harder to manipulate. An example of this is the output value variable, you don't really need it there. Think of it this way: if it doesn't need to persist between game modes, then it's probably a good idea to not put them in the global scope.
- I've seen a couple of unnecessary variables here, like say, on line 53 and 63, you're declaring the `numberArray` variable and assigning the value of `player` array to it, in this instance, why not just use the player array itself? why is there a need to declare the numberArray variable?
- Another example here is on the switch cases, you're watching the `numOfPlayers` variable for a change. But this is totally unnecessary, as what is the number of players? It can also be represented as `players.length`. As you are popping the players out one by one, it's going to change the length as well.

All that being said, your work is still great, Diron! I can't wait to see what you are able to do for the blackjack project!

PR #471, opened 2023-02-23
i almost gave up 20-1 jiajun

legenthepigxD:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

6 hours
input not defined
concatenating the numbers
Comfort Level (1-5):
3
Completeness Level (1-5):
4
it was alright, i had to pull up the walkthrough, but i ended up learning so much more by doing so.
switching gamemodes was very fun
nope, thank you!

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,82 @@
+var gamemode = 1;
+var player1Dice = [];
+var player2Dice = [];
+var player1Wc = "";
+var player2Wc = "";
+
+var diceRoller = function () {
+  var roll = Math.random() * 6;
+  var roundDown = Math.floor(roll) + 1;
+  return roundDown;
+};
+var newDice = function () {
+  var newDices = [diceRoller(), diceRoller()];
+  if (gamemode == 1) {
+    player1Dice = newDices;
+  } else if (gamemode == 3) {
+    player2Dice = newDices;
+  }
+  console.log(newDices);
+  return newDices;
+};
+
+var concatenateNumbers = function (num1, num2) {
+  return Number(String(num1) + String(num2));
+};
+
+var playerChoice = function (input) {
+  if (gamemode == 2) {
+    if (input == `a`) {
+      player1Num = concatenateNumbers(player1Dice[0], player1Dice[1]);
+    } else if (input == `b`) {
+      player1Num = concatenateNumbers(player1Dice[1], player1Dice[0]);
+    }
+    player1Wc = player1Num;
+    return player1Num;
+  }
+};
+var playerChoice2 = function (input) {
+  if (gamemode == 4) {
+    if (input == `a`) {
+      player2Num = concatenateNumbers(player2Dice[0], player2Dice[1]);
+    } else if (input == `b`) {
+      player2Num = concatenateNumbers(player2Dice[1], player2Dice[0]);
+    }
+    player2Wc = player2Num;
+    return player2Num;
+  }
+};
+
+var decider = function (input) {
+  if (player1Wc > player2Wc) {
+    return 1;
+  } else return 2;
+};
 var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
+  var dice = newDice();
+  if (gamemode == 1) {
+    var msg = `Hi player 1!<br/> The dices rolled ${player1Dice[0]} (a) and ${player1Dice[1]} (b) <br/> <br/> please choose which number will be the first digit to the combined number <br/><br/> you can do so by inputting a or b`;
+    gamemode = 2;
+  } else if (gamemode == 2) {
+    msg = `player 1's number is ${playerChoice(input)}`;
+    gamemode = 3;
+  } else if (gamemode == 3) {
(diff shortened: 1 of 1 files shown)


script.js
```diff
@@ -1,4 +1,82 @@
...
+  if (gamemode == 1) {
+    player1Dice = newDices;
+  } else if (gamemode == 3) {
+    player2Dice = newDices;
+  }
+  console.log(newDices);
+  return newDices;
+};
+
+var concatenateNumbers = function (num1, num2) {
```
floatingtales, line 23: I like how you have this in a separate function! Good job!

script.js
```diff
@@ -1,4 +1,82 @@
...
+    if (input == `a`) {
+      player1Num = concatenateNumbers(player1Dice[0], player1Dice[1]);
+    } else if (input == `b`) {
+      player1Num = concatenateNumbers(player1Dice[1], player1Dice[0]);
+    }
+    player1Wc = player1Num;
+    return player1Num;
+  }
+};
+var playerChoice2 = function (input) {
```
floatingtales, line 38: technically you can merge this with the previous function, as the contents are somewhat the same, and you can just add a small if statement to see which player is assigned the number

script.js
```diff
@@ -1,4 +1,82 @@
...
+    }
+    player2Wc = player2Num;
+    return player2Num;
+  }
+};
+
+var decider = function (input) {
+  if (player1Wc > player2Wc) {
+    return 1;
+  } else return 2;
```
floatingtales, line 53: Yep, there is the possibility where you would find players have the same numbers, I know it's an extra step to add, but these are what we can call "fringe cases" and it should be considered.

script.js
```diff
@@ -1,4 +1,82 @@
...
+};
 var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
+  var dice = newDice();
```
floatingtales, line 56: having this function call here, means that every single time you clicked on the button, dice is rolled, yet it's not used unless it's in gamemode 1 or gamemode 3, perhaps it's better to have this on gamemode 1 and 3 only

script.js
```diff
@@ -1,4 +1,82 @@
...
+    msg = `player 1's number is ${playerChoice(input)}`;
+    gamemode = 3;
+  } else if (gamemode == 3) {
+    msg = `Hi player 2, now it's your turn!<br/>The dices rolled ${player2Dice[0]} (a) and ${player2Dice[1]} (b) <br/> <br/> please choose which number will be the first digit to the combined number <br/><br/> you can do so by inputting a or b`;
+    gamemode = 4;
+  } else if (gamemode == 4) {
+    msg = `player 2's number is ${playerChoice2(input)}`;
+    gamemode = 5;
+  } else if (gamemode == 5) {
+    var winner = decider();
```
floatingtales, line 70: this is a great way to find out who wins! This is good, buddy!

script.js
```diff
@@ -1,4 +1,82 @@
...
+    msg = `player ${winner} wins!`;
+    gamemode = 0;
+  } else if (gamemode == 0) {
```
floatingtales, line 73: a nice reset case here! This means that you can play the game over and over again!

script.js
```diff
@@ -1,4 +1,82 @@
...
+    player1Dice = [];
+    player2Dice = [];
+    gamemode = 1;
+    msg = `press submit to play again!`;
+  }
+
+  return msg;
+  ``;
```
floatingtales, line 81: this final string literal is unnecessary

script.js
```diff
@@ -1,4 +1,82 @@
...
+  if (player1Wc > player2Wc) {
+    return 1;
+  } else return 2;
+};
 var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
+  var dice = newDice();
+  if (gamemode == 1) {
+    var msg = `Hi player 1!<br/> The dices rolled ${player1Dice[0]} (a) and ${player1Dice[1]} (b) <br/> <br/> please choose which number will be the first digit to the combined number <br/><br/> you can do so by inputting a or b`;
```
floatingtales, line 58: This is not good practice, the variable `msg` is declared here in the if code block, hence, it should only exist within lines 57-60, because of its scope, yet you're using this variable for all the subsequent game modes, and for the output. This only works because it's JavaScript, and there's this concept called 'variable hoisiting' where a variable is 'hoisted' up to the scope where it's needed, but this is a JavaScript only property, and an outdated one at that.

floatingtales (commented):
This was a great attempt on the project, Jia Jun! It's fairly simple, and straightforward, and I think you're grasping the concepts well. I don't have a lot of notes, but I do have a few particular nitpick:

- It seemed that I see there are quite a few repeating codes, you have two functions that basically do the same thing for each player, and you have modes with output messages similar to one another, it would be best if you can find ways of optimizing the code by combining them together, somewhat.
- Please do not forget to declare variables in their respective scope, buddy! This will not work if it was coded in any other languages.

But other than these few things, you've done a great job on this, buddy! I can't wait to see what you can do in your blackjack project!

PR #492, opened 2023-04-13
testing 1

calvinttg:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

3

 (var currentPlayerRollDice = [];) I redeclared a variable which causes the array to be cleared instead of holding the values 

Second player logic 

Comfort Level (1-5):
3
Completeness Level (1-5):
3

-

Able to find to the specific issue that has been giving me headaches which is everything

Using the Number() to convert String value to numerals and using .push to add it into the array. Makes it feel like i am automating a process

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,152 @@
-var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
+//set gameState to gameStateRollDice
+//Players rolls 2 dice - gameStateRollDice
+//Players chooses dice orientation -gameStateChooseDice
+//Prompt to choose between 1 & 2, will show error msg if input is not 1 or 2
+//Add dice number as string(to not add number) then convert to numerals using Number()
+//Compare combined numbers accross all players and higher combined number wins
+// Add score to winner
+// Press submit button to roll dice (set gameState to gameStateRollDice)
+
+var gameStateRollDice = "gamestate is roll dice";
+var gameStateChooseDice = "game state is choose dice";
+var gameStateCompareScore = "game state is comparing Scores";
+var gameState = gameStateRollDice;
+
+var currentPlayerDiceNumber = [];
+
+var currentPlayer = 1;
+var scoreBoard = [];
+
+var rollDice = function () {
+  // Generate a decimal from 0 through 6, inclusive of 0 and exclusive of 6.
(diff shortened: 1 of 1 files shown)


script.js
```diff
@@ -1,4 +1,152 @@
...
+
+var currentPlayer = 1;
+var scoreBoard = [];
+
+var rollDice = function () {
+  // Generate a decimal from 0 through 6, inclusive of 0 and exclusive of 6.
+  var randomDecimal = Math.random() * 6;
+  // Remove the decimal with the floor operation.
+  // This will be an integer from 0 to 5 inclusive.
+  // Add 1 to get valid dice rolls of 1 through 6 inclusive.
```
hpgay, line 25: Really appreciate the comments

script.js
```diff
@@ -1,4 +1,152 @@
...
+    console.log(scoreBoard);
+    console.log(currentPlayerDiceNumber);
+    console.log("currentplayer2score: " + playerScore);
+    return "Your score is: " + playerScore;
+  }
+};
+
+var comparePlayerScore = function () {
+  var compareMessage =
+    "Player 1 score: " + scoreBoard[0] + "<br>Player 2 score: " + scoreBoard[1];
```
hpgay, line 90: Can consider initiating variables like this:
```
var player1Score = scoreBoard[0];
var player2Score = scoreBoard[1];
```

and then use these variables later on in this function instead of directly `scoreBoard[0/1]` just because it's not as intuitive what it might be referring to.

script.js
```diff
@@ -1,4 +1,152 @@
...
+      return (
+        myOutputMessage +
+        "<br> Now is Player 2's turn. <br>Please enter 1 or 2 to choose corresponding dice"
+      );
+    }
+    if (currentPlayer == 2) {
+      console.log("End of Player 2's turn. Calculating scores");
+      gameState = gameStateCompareScore;
+      return myOutputMessage + "<br>Click submit to calculate score.";
+    }
```
hpgay, line 135: Abstract the code for each gamestate into their own function

hpgay (commented):
Good functions and naming, can decompose the code further into their own function so that you have less clutter in the main function.

PR #518, opened 2023-07-24
23-2 Iffah Beat that

IffahA:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

How many hours did you spend on this assignment? 10hr

 Variable no found
debugging

Comfort Level (1-5):
2 found it hard at first but when I broke down the different functions it helped to see how to create the game

Completeness Level (1-5):
4

The game works based on my test so happy about that

Is there anything in this code that you feel pleased about? That I found a way to continue if user failed validation without asking them to start from beginning

anywhere I can refactor further

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,149 @@
+// Requirements
+// 1. There are 2 players and players take turns.
+// 2. When a player clicks Submit, the game rolls 2 dice and shows the dice rolls, for example 3 and 6.
+// 3. The player picks the order of the dice they want. For example, if they wanted the number 63, they would specify that the 2nd dice goes first. You can choose how the player specifies dice order.
+// 4. After both players have rolled and chosen dice order, the player with the higher combined number wins.
+
+//Define global variables
+//Checks if game mode is to roll dice or if user needs to specify order
+var GAME_MODE_ROLL_DICE = `dice Roll`;
+var GAME_MODE_CHOOSE_ORDER = `choose Order`
+var UNIVERSAL_GAME_MODE = GAME_MODE_ROLL_DICE;
+var counter = 0;
+var firstPlayerFinalDiceRoll = 0;
+var secondPlayerFinalDiceRoll = 0;
+var playerDiceRoll = [];
+var errorMessage = ``;
+var chosenOrderOfDiceRoll = 0;
+
+
 var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
-};
+
+  //loop exits after 2 rounds where player 1 & 2 choose dice values
+  while (counter < 2) {
+  // starts game with dice roll
+  if (UNIVERSAL_GAME_MODE === GAME_MODE_ROLL_DICE) { 
+    var totalDiceRolls = [];
+    totalDiceRolls.push(rollDice());
+    UNIVERSAL_GAME_MODE = GAME_MODE_CHOOSE_ORDER;
+    return totalDiceRolls;
+  }
+  //proceeds to game section to choose dice order
+  if ( UNIVERSAL_GAME_MODE === GAME_MODE_CHOOSE_ORDER ) {
+    chosenOrderOfDiceRoll = validatesInput(input);
+
+    if (chosenOrderOfDiceRoll === errorMessage) {
+      return chosenOrderOfDiceRoll;
+    }
+
+    if (chosenOrderOfDiceRoll !== errorMessage) {
+      var reorderDiceRoll = orderDiceRoll(chosenOrderOfDiceRoll);
+      var finalDiceOrder = capturesDiceRollForEachPlayer(reorderDiceRoll);
+      UNIVERSAL_GAME_MODE = GAME_MODE_ROLL_DICE
+      return finalDiceOrder;
+    }
+  }
+    }
+    // evaluates whether player 1 or 2 wins based on higher dice value
+ var myOutputValue = testWinningCondition();
+ return myOutputValue;
+  } 
+
+//Function to roll dice
+var generateDiceRoll = function() {
+  var randomDecimal = Math.random() * 6;
+  var randomInteger = Math.floor( randomDecimal );
+  var diceRoll = randomInteger + 1;
+  return diceRoll;
+}
+
+//Function to roll dice twice
+var rollDice = function() {
+  var rollDiceCounter = 0;
+  playerDiceRoll.length = 0;
+  while (rollDiceCounter < 2) {
+    var newDiceRoll = generateDiceRoll();
+     playerDiceRoll.push(newDiceRoll);
+    rollDiceCounter += 1;
+   }
+
+    var myOutputValue = `Welcome! <br> You rolled ${playerDiceRoll[0]} for Dice 1 and ${playerDiceRoll[1]} for Dice 2.  <br> Choose the order of the dice.- Enter 1 to start with Dice 1: ${playerDiceRoll[0]} or 2 to start with Dice 2:${playerDiceRoll[1]}.`
+    return myOutputValue;
+}
+
+//Function to validate input; whether user entered 1 or 2
+var validatesInput = function(input) {
+  if ( (input !== `1`) && (input !== `2`) ) {
+    errorMessage = `Wrong option! Please enter either 1 or 2 only.`;
+    return errorMessage;
+  } else {
+    chosenOrder = input;
+    return chosenOrder;
+    
+  }
+}
+
+//Function to reorder dice roll based on user input
+var orderDiceRoll = function(input) {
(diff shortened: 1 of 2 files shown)


script.js
```diff
@@ -1,4 +1,149 @@
+// Requirements
+// 1. There are 2 players and players take turns.
+// 2. When a player clicks Submit, the game rolls 2 dice and shows the dice rolls, for example 3 and 6.
+// 3. The player picks the order of the dice they want. For example, if they wanted the number 63, they would specify that the 2nd dice goes first. You can choose how the player specifies dice order.
+// 4. After both players have rolled and chosen dice order, the player with the higher combined number wins.
+
+//Define global variables
+//Checks if game mode is to roll dice or if user needs to specify order
+var GAME_MODE_ROLL_DICE = `dice Roll`;
+var GAME_MODE_CHOOSE_ORDER = `choose Order`
```
mahe993, line 10: good use of constants, but we don't have to camelCase the actual string values. They are actual values and should be cased however the string should look like

script.js
```diff
@@ -1,4 +1,149 @@
...
+  playerDiceRoll.length = 0;
+  while (rollDiceCounter < 2) {
+    var newDiceRoll = generateDiceRoll();
+     playerDiceRoll.push(newDiceRoll);
+    rollDiceCounter += 1;
+   }
+
+    var myOutputValue = `Welcome! <br> You rolled ${playerDiceRoll[0]} for Dice 1 and ${playerDiceRoll[1]} for Dice 2.  <br> Choose the order of the dice.- Enter 1 to start with Dice 1: ${playerDiceRoll[0]} or 2 to start with Dice 2:${playerDiceRoll[1]}.`
+    return myOutputValue;
+}
```
mahe993, line 72: ideally, a function should do what its name suggests. Just by looking at the name `rollDice`, i would assume this is rolling one dice and returning that dice value and would not expect this function to be returning a string

script.js
```diff
@@ -1,4 +1,149 @@
...
-  var myOutputValue = 'hello world';
-  return myOutputValue;
-};
+
+  //loop exits after 2 rounds where player 1 & 2 choose dice values
+  while (counter < 2) {
+  // starts game with dice roll
+  if (UNIVERSAL_GAME_MODE === GAME_MODE_ROLL_DICE) { 
+    var totalDiceRolls = [];
+    totalDiceRolls.push(rollDice());
```
mahe993, line 27: if `rollDice()` returns us an output string directly, we do we have to store it in an array before returning it?

script.js
```diff
@@ -1,4 +1,149 @@
...
+var chosenOrderOfDiceRoll = 0;
+
+
 var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
-};
+
+  //loop exits after 2 rounds where player 1 & 2 choose dice values
+  while (counter < 2) {
```
mahe993, line 23: the while loop is essentially being used as a `if` condition here. `counter` is also being used for too many use cases, i would suggest to keep a specific variable for exiting a loop and add to that variable explicitly within the end of the loop rather than using a global variable.

script.js
```diff
@@ -1,4 +1,149 @@
...
+    if (chosenOrderOfDiceRoll === errorMessage) {
+      return chosenOrderOfDiceRoll;
+    }
+
+    if (chosenOrderOfDiceRoll !== errorMessage) {
+      var reorderDiceRoll = orderDiceRoll(chosenOrderOfDiceRoll);
+      var finalDiceOrder = capturesDiceRollForEachPlayer(reorderDiceRoll);
+      UNIVERSAL_GAME_MODE = GAME_MODE_ROLL_DICE
+      return finalDiceOrder;
+    }
```
mahe993, line 44: if `chosenOrderOfDiceRoll` is equal `errorMessage`, we already return. Whatever follows must naturally mean `chosenOrderOfDiceRoll !== errorMessage` so we do not have to explicitly check for it.
```suggestion
    if (chosenOrderOfDiceRoll === errorMessage) {
      return chosenOrderOfDiceRoll;
    }

    var reorderDiceRoll = orderDiceRoll(chosenOrderOfDiceRoll);
    var finalDiceOrder = capturesDiceRollForEachPlayer(reorderDiceRoll);
    UNIVERSAL_GAME_MODE = GAME_MODE_ROLL_DICE
    return finalDiceOrder;
    
```

script.js
```diff
@@ -1,4 +1,149 @@
...
+  var randomDecimal = Math.random() * 6;
+  var randomInteger = Math.floor( randomDecimal );
+  var diceRoll = randomInteger + 1;
+  return diceRoll;
+}
+
+//Function to roll dice twice
+var rollDice = function() {
+  var rollDiceCounter = 0;
+  playerDiceRoll.length = 0;
```
mahe993, line 63: did you meant to check this or are you trying to assign a length? a length cannot/should not be assigned a value as it is simply checking the length of the array. If the intended action is to reset the array back to empty, you can simply `playerDiceRoll = []`

script.js
```diff
@@ -1,4 +1,149 @@
...
+    var myOutputValue = `Welcome! <br> You rolled ${playerDiceRoll[0]} for Dice 1 and ${playerDiceRoll[1]} for Dice 2.  <br> Choose the order of the dice.- Enter 1 to start with Dice 1: ${playerDiceRoll[0]} or 2 to start with Dice 2:${playerDiceRoll[1]}.`
+    return myOutputValue;
+}
+
+//Function to validate input; whether user entered 1 or 2
+var validatesInput = function(input) {
+  if ( (input !== `1`) && (input !== `2`) ) {
+    errorMessage = `Wrong option! Please enter either 1 or 2 only.`;
+    return errorMessage;
+  } else {
```
mahe993, line 79: do we need the `else`?

script.js
```diff
@@ -1,4 +1,149 @@
...
+    return chosenOrder;
+    
+  }
+}
+
+//Function to reorder dice roll based on user input
+var orderDiceRoll = function(input) {
+  //if user chose order of dice 1 then dice 2
+  if (input === `1`) {
+    var finalOrder = Number( `` + playerDiceRoll[0] + playerDiceRoll[1] );
```
mahe993, line 90: interesting use of `` to turn the dice rolls into string, although its not something i would advise because it seems like a hacky and unintended way to do it. Is there a better way we can do this? maybe just through numbers:
```
var finalOrder = playerDiceRoll[0] * 10 + playerDiceRoll[1]
```

script.js
```diff
@@ -1,4 +1,149 @@
...
+  counter += 1;
+
+//initialize final dice value for player 1
+  if (( UNIVERSAL_GAME_MODE === GAME_MODE_CHOOSE_ORDER ) && (counter === 1) ) {
+      firstPlayerFinalDiceRoll = input;
+      var myOutputValue = `The first player's dice roll is ${firstPlayerFinalDiceRoll}. Please click submit again to roll for second player.`
+      return myOutputValue;
+    }
+    //initialize final dice value for player 2
+    else if (( UNIVERSAL_GAME_MODE === GAME_MODE_CHOOSE_ORDER ) && (counter === 2 ) ) {
```
mahe993, line 111: Lookout for the `else` !

script.js
```diff
@@ -1,4 +1,149 @@
...
+    var myOutputValue = `Player 1 wins! Player 1 rolled ${firstPlayerFinalDiceRoll} while player 2 rolled ${secondPlayerFinalDiceRoll}.`
+
+     //reset counter & game mode
+    counter = 0;
+    UNIVERSAL_GAME_MODE = GAME_MODE_ROLL_DICE;
+
+    return myOutputValue;
+  }
+  //if 2nd player dice roll is more than 1st player
+  else if (secondPlayerFinalDiceRoll > firstPlayerFinalDiceRoll) {
```
mahe993, line 131: `else`

script.js
```diff
@@ -1,4 +1,149 @@
...
+    var myOutputValue = `Player 2 wins! Player 2 rolled ${secondPlayerFinalDiceRoll} while player 1 rolled ${firstPlayerFinalDiceRoll}.`
+
+     //reset counter & game mode
+    counter = 0;
+    UNIVERSAL_GAME_MODE = GAME_MODE_ROLL_DICE;
+
+    return myOutputValue;
+  }
+  //the only other condition left is if they tie
```
mahe993, line 140: this shows you understand implicity! try to apply this concept to other if else statements

mahe993:
overall, great job on a working product! remember that the end user doesn't care how your code looks like, they care about how simple/fun/easy it is to use your app! Having a working product is commendable in itself! Keep up the great work!

PR #519, opened 2023-07-24
23-2_Ignatius Tan 

ignatiusgabrieltan:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

5-6 hours

I got a few undefines.

figuring out how the motion goes from player 1 to player 2 to the scoreboard.

Comfort Level (1-5):
5

Completeness Level (1-5):
5

It's a fun game to make which made me use all the tools I've learnt so far.

Using more of the HTML aspect to make the layout look nicer.

To learn more HTML so that I can do "super comfortable".

### DIFF
--- a/index.html
+++ b/index.html
@@ -18,7 +18,7 @@
       }
 
       #container {
-        background-color: pink;
+        background-color: #ff69b4;
         margin: 40px auto;
         max-width: 800px;
         padding: 38px 31px;
@@ -33,30 +33,52 @@
       }
 
       button {
-        font-size: 21px;
-        line-height: 33px;
-        margin: 0 0 23px 0;
-        padding: 0 6px;
+        display: block;
+        width: 80%;
+        margin: 1em auto;
+        height: 3em;
+        font-size: 1.1rem;
+        min-width: 300px;
       }
 
       #output-div {
-        background-color: lightgrey;
+        background-color: whitesmoke;
         margin: 20px auto;
         padding: 20px;
         width: 100%;
+      }
+
+      h3 {
+        display: block;
+        color: navy;
+        margin: 20px auto;
+        font-size: 1.1rem;
       }
     </style>
   </head>
 
   <body>
-    <h1 id="header">Basics: Beat That! 🚀</h1>
+    <h1 id="header">🎲⛹️ Beat That! 🎲⛹️</h1>
     <div id="container">
       <p>Input:</p>
       <input id="input-field" />
       <br />
       <button id="submit-button">Submit</button>
       <p>Output:</p>
-      <div id="output-div"></div>
+      <div id="output-div">
+        <p>Welcome to...BEAT THAT!</p>
+        <p>Please press submit to start the game!</p>
+      </div>
+      <h3>
+        <center>
+          Welcome to Beat That!<br />This is a 2 player game(player1 &
+          player2).<br />You both will compete with each other to see who can<br />
+          make up the most points using 2 dice rolls.<br />You can select which
+          number to go first.<br />e.g You've roll 1 & 6<br />By typing "1" or
+          "2" and submiting will<br />determine which number goes first.<br />typing
+          "1" gives you 16<br />typing "2" gives you 61<br />🤼‍♂️LET'S PLAY!!!🤼‍♂️
+        </center>
+      </h3>
     </div>
     <!-- Import program logic -->
     <script src="script.js"></script>
--- a/script.js
+++ b/script.js
@@ -29,10 +29,13 @@ var main = function (input) {
   }
   if (GAMEMODE === `PLAYER2 CHOICE`) {
     myOutputValue = whenPlayer2ChoosesTheOrder(input);
+    return myOutputValue;
   }
 
   //Score board
   if (GAMEMODE === `SCOREBOARD`) {
+    myOutputValue = scoreBoard();
+    return myOutputValue;
   }
   return myOutputValue;
 };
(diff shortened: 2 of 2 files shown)


index.html
```diff
@@ -18,7 +18,7 @@
       }
 
       #container {
-        background-color: pink;
+        background-color: #ff69b4;
```
mahe993, line 21: minus points for this pink it burns!!! 🤣

script.js
```diff
@@ -1,4 +1,171 @@
+//GLOBAL VARIABLES
+var GAMEMODE = `PLAYER1`;
+
+// Arrays for players 1 and 2
+var PLAYER1ARRAY = [];
+var PLAYER2ARRAY = [];
```
mahe993, line 6: arrays are never constants because they are mutable data structures (you can mutate them by push/pop/manipulate its value etc. So they should never be uppercased like a string constant

script.js
```diff
@@ -1,4 +1,171 @@
...
+  return diceNumber;
+};
+
+//Functions when it's player1s turn...
+
+//When player1 rolls the dice...
+var whenPlayer1Rolls = function (input) {
+  //Random Dice rolls
+  diceRoll01 = rollDice();
+  diceRoll02 = rollDice();
```
mahe993, line 63: `diceRoll01` and `diceRoll02` are not declared anywhere. Why are they needed in the first place?

script.js
```diff
@@ -1,4 +1,171 @@
...
+    return myOutputValue;
+  }
+
+  //Score board
+  if (GAMEMODE === `SCOREBOARD`) {
+    myOutputValue = scoreBoard();
+    return myOutputValue;
+  }
+  return myOutputValue;
+};
```
mahe993, line 41: as always, 😀👍👍👍

mahe993: This will go a loooong way in helping you next time. You eventually learn not just to refactor into its own function, but as you start noticing such patterns you will be able to make the function reusable across different scenarios!

script.js
```diff
@@ -1,4 +1,171 @@
...
+    GAMEMODE = `PLAYER2`;
+
+    // add value into PLAYER1 ARRAY for future scores...
+    PLAYER1ARRAY.push(rolledDice01 * 10 + rolledDice02);
+    console.log(PLAYER1ARRAY);
+    return `Player1! You've choosen ${rolledDice01}${rolledDice02}!<br>Now it's Player2s' turn to roll the dice.<br>
+  press submit to continue.`;
+  }
+
+  if (order === "2") {
```
mahe993, line 91: because you already checked `if (order !== "1" && order !== "2")` earlier, the value of `order` at this point forth must be either 1 or 2. once you have checked `if (order === "1")`, it must mean whatever is left must be 2, so we do not need ` if (order === "2")`

script.js
```diff
@@ -1,4 +1,171 @@
...
+    GAMEMODE = `PLAYER2`;
+
+    // add value into PLAYER1 ARRAY for future scores...
+    PLAYER1ARRAY.push(rolledDice02 * 10 + rolledDice01);
```
mahe993, line 95: 👍

script.js
```diff
@@ -1,4 +1,171 @@
...
+
+//When player2 rolls the dice...
+var whenPlayer2Rolls = function (input) {
+  //Random Dice rolls
+  diceRoll01 = rollDice();
+  diceRoll02 = rollDice();
+
+  //rolled Dice imprints the random dice rolls...
+  rolledDice01 = diceRoll01;
+  rolledDice02 = diceRoll02;
```
mahe993, line 113: same as above

script.js
```diff
@@ -1,4 +1,171 @@
...
+  if (order === "2") {
+    GAMEMODE = `SCOREBOARD`;
+
+    // add value into PLAYER1 ARRAY for future scores...
+    PLAYER2ARRAY.push(rolledDice02 * 10 + rolledDice01);
+
+    return `Player2! You've choosen ${rolledDice02}${rolledDice01}!<br>Now lets see who has won!<br>
+  press submit to continue.`;
+  }
+};
```
mahe993, line 144: this is essentially an exact same func as `whenPlayer1ChoosesTheOrder`, maybe we can try making this reusable across both cases

script.js
```diff
@@ -1,4 +1,171 @@
...
+
+//SCOREBOARD
+var scoreBoard = function (input) {
+  //score for Player 1
+  var PLAYER1sScore = 0;
```
mahe993, line 149: camelCasing please!

script.js
```diff
@@ -1,4 +1,171 @@
...
+  PLAYER2ARRAY.forEach((item) => {
+    PLAYER2sScore += item;
+  });
+
+  if (PLAYER1sScore > PLAYER2sScore) {
+    GAMEMODE = `PLAYER1`;
+    myOutputValue = `PLAYER 1 WINS!<br>Here are the scores!<br>PLAYER 1 - ${PLAYER1sScore}<BR>PLAYER 2 - ${PLAYER2sScore}<br>Press submit to continue!`;
+  }
+  if (PLAYER1sScore < PLAYER2sScore) {
+    GAMEMODE = `PLAYER1`;
```
mahe993, line 167: we can use a else if here so that the program doesnt continue checking unnecessarily. in both scenarios, we also do `GAMEMODE = "PLAYER1`, mayb we can do that one time after this two `if`s happen

mahe993:
Very nice clean code, reviewing this is a breeze! Thanks for that! Very well done, except for the pink!

PR #523, opened 2023-07-24
Basics-Beat-That Submission: Christine Ravindaran

ChrisR3893:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

How many hours did you spend on this assignment? **4 hours**

Please fill in one error and/or error message you received while working on this assignment. **I failed to convert the result from the dice rolls into a string to concatenate before casting it as an integer. Instead, I added up the numbers from the dice rolls which resulted in the wrong result.**

What part of the assignment did you spend the most time on? **Determining the flow of the different steps involving players 1 and 2 and determining how many global variables were needed.**

Comfort Level (1-5): **4**

Completeness Level (1-5): **4**

What did you think of this deliverable? **I found it manageable as a whole. It is also quite similar to some of the in-class exercises.**

Is there anything in this code that you feel pleased about? **The use of comments/pseudocode to clearly illustrate the logic used.**

What's one aspect of your code you would like specific, elaborate feedback on? **The part on whether it could be more concise with the introduction of helper functions or good coding practices.**

### DIFF
--- a/script.js
+++ b/script.js
@@ -1,4 +1,99 @@
+// BEAT THAT PROJECT Requirements
+// There are 2 players and players take turns
+// When a player clicks Submit, the game rolls 2 dice and shows the dice rolls, for example 3 and 6.
+// The player picks the order of the dice they want. For example, if they wanted the number 63, they would specify that the 2nd dice goes first. //
+// You can choose how the player specifies dice order.
+// After both players have rolled and chosen dice order, the player with the higher combined number wins.
+
+// Introduce global variable to distinguish between game, input entry and results modes
+var gameMode = "awaiting user input";
+
+// Introduce global variable to distinguish between player 1 and player 2 modes
+var playerMode = "no player";
+
+// Define global variables to hold values
+var diceRoll1 = 0;
+var diceRoll2 = 0;
+var player1Num = 0;
+var player2Num = 0;
+
+// Function that gets executed when the submit button is clicked
 var main = function (input) {
-  var myOutputValue = 'hello world';
-  return myOutputValue;
+  // Case where new game round starts with Player 1
+  if (gameMode === "awaiting user input" && playerMode === "no player") {
+    diceRoll1 = rollDice();
+    diceRoll2 = rollDice();
+    myOutput = `Welcome Player 1. You rolled ${diceRoll1} for Dice 1 and ${diceRoll2} for Dice 2.<br>Choose the order of the dice by typing in 12 or 21.<br>12 denotes that the first dice comes before the second and 21 denotes vice versa.`;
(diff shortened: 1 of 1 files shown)


script.js
```diff
@@ -1,4 +1,99 @@
...
+  else if (gameMode === "awaiting user input" && playerMode === "Player 2") {
+    diceRoll1 = rollDice();
+    diceRoll2 = rollDice();
+    myOutput = `Welcome Player 2. You rolled ${diceRoll1} for Dice 1 and ${diceRoll2} for Dice 2.<br>Choose the order of the dice by typing in 12 or 21.<br>12 denotes that the first dice comes before the second and 21 denotes vice versa.`;
+    gameMode = "game time";
+    playerMode = "Player 2";
+  } // Case after Player 2 provides the order
+  else if (gameMode === "game time" && playerMode === "Player 2") {
+    if (input === "12") {
+      player2Num = Number(String(diceRoll1) + String(diceRoll2));
```
mahe993, line 56: Ideally, I would not like you to make numbers into strings just to concatenate them and turn them back into numbers. this can be achieved by:
```
diceRoll1 * 10 + diceRoll2
```

script.js
```diff
@@ -1,4 +1,99 @@
...
+      gameMode = "awaiting user input";
+      return "Invalid input, the dice will roll again and please type in either 12 or 21.";
+    }
+  } // Case where game continues with Player 2
+  else if (gameMode === "awaiting user input" && playerMode === "Player 2") {
+    diceRoll1 = rollDice();
+    diceRoll2 = rollDice();
+    myOutput = `Welcome Player 2. You rolled ${diceRoll1} for Dice 1 and ${diceRoll2} for Dice 2.<br>Choose the order of the dice by typing in 12 or 21.<br>12 denotes that the first dice comes before the second and 21 denotes vice versa.`;
+    gameMode = "game time";
+    playerMode = "Player 2";
```
mahe993, line 52: why do we need to do this? we are only able to reach this part of the code if `playerMode` is already 2

mahe993:
I think one of the main issues here is that everything is clogged up in `main`, so some refactoring could be done here if you have the capacity for it. The use of the `if`s and `else if`s are also very verbose and difficult to follow. A pattern that is instantly recognizable is that each larger `if` is checking firstly game mode, and player turn. Maybe we can do something like:
```
if (gameMode === someMode) {
 if (playerMode === 1) {
   return playSomeMode(someValue, player1)
 } 
 return playSomeMode(someValue, player2)
}

if (gameMode === nextGameMode) {
...
}
```

PR #526, opened 2023-07-25
Marina, project 2: beat that!

Mannushka:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

How many hours did you spend on this assignment? 9-10. I first tried to do it on my own without watching the video to see what I am able to come up with :D I finished first three stages but my code was kinda heavy and primitive (i had 2 variables for each player's score, didn't use the loop for dice roll so had just 2 variables "dice1" and "dice2" etc.), so I rewrote it relying on the tutorial.

My errors had to do with conditionals mostly, they wouldn't run. 
I struggled a bit with input validation for the dice order, I knew what was wrong but it took me awhile to come up with an idea how to fix it.

Fixing errors in input validation and other functions (already forgot which ones precisely).

Comfort Level (1-5): 3, may be even 4 with the tutorial.

Completeness Level (1-5): 4

It works as intended.
Anything that looks strange/illogical/inconsistent.

### DIFF
--- a/script.js
+++ b/script.js
@@ -18,6 +18,7 @@ var currentPlayer = 1;
 var currentPlayerRolls = []; //this array will store the currrent player rolls
 var allPlayersScore = [];
 var playerFinalValue = 0;
+var errorMessage = "";
 //Helper function "Roll the dice"
 var rollDice = function () {
   var randomDecimal = Math.random() * 6;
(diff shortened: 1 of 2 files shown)


script.js
```diff
@@ -1,4 +1,165 @@
...
+
+//1) a) create a dice game with 2 dices for player 1, b) let player 1 to choose which dice they want to be the first one.
+//2) include player 2 and let them do same things as player 1
+//3)compare the 2 dice rolls and announce the winner
+
+//Global variables:
+var GAME_STATE_ROLL_DICE = "GAME_STATE_ROLL_DICE ";
+var GAME_STATE_CHOOSE_DICE_ORDER = "GAME_STATE_CHOOSE_DICE_ORDER";
+var GAME_STATE_COMPARE_SCORES = "GAME_STATE_COMPARE_SCORES";
+var GAME_STATE_PLAY_AGAIN = "GAME_STATE_PLAY_AGAIN";
```
mahe993, line 15: good use of string constants! the value itself does not need to be upper cased unlike the variable naming since it is afterall, a string.

script.js
```diff
@@ -1,4 +1,165 @@
...
+  //input validation:
+  if (input != 1 && input != 2) {
+    errorMessage =
+      "Error! Please only key in 1 or 2 to choose which dice to use as the first digit. Your current dice rolls are " +
+      currentPlayerRolls[0] +
+      " as dice 1 and " +
+      currentPlayerRolls[1] +
+      " as dice 2.";
+    return errorMessage;
+  } else {
```
mahe993, line 59: unnecessary `else` here

script.js
```diff
@@ -1,4 +1,165 @@
...
+        `${currentPlayerRolls[0]}` + `${currentPlayerRolls[1]}`
+      );
+    }
+
+    //condition 2 "if input == 2"
+    if (input == 2) {
+      playerFinalValue = Number(
+        `${currentPlayerRolls[1]}` + `${currentPlayerRolls[0]}`
+      );
+    }
```
mahe993, line 72: this is where `else` would come in handy
```suggestion
    if (input == 1) {
      playerFinalValue = Number(
        `${currentPlayerRolls[0]}` + `${currentPlayerRolls[1]}`
      );
    } else {
        //condition 2 "if input == 2"
      playerFinalValue = Number(
        `${currentPlayerRolls[1]}` + `${currentPlayerRolls[0]}`
      );
    }
```

script.js
```diff
@@ -1,4 +1,165 @@
...
+    return (
+      "Player's 1 final value is " +
+      allPlayersScore[0] +
+      ". <br> <br> Player's 2 final value is " +
+      allPlayersScore[1] +
+      ". <br> <br> " +
+      "Player 1 wins."
+    );
+    //player 2 wins
+  } else if (allPlayersScore[1] > allPlayersScore[0]) {
```
mahe993, line 97: `else` is not necessary here, we can keep the `if`!

script.js
```diff
@@ -1,4 +1,165 @@
...
+    return (
+      "Player's 1 final value is " +
+      allPlayersScore[0] +
+      ". <br> <br> Player's 2 final value is " +
+      allPlayersScore[1] +
+      ". <br> <br> " +
+      "Player 2 wins."
+    );
+    // a tie:
+  } else {
```
mahe993, line 107: `else` is not necessary, whatever remains here must be `allPlayersScore[0] === allPlayersScore[1]` implicitly

script.js
```diff
@@ -1,4 +1,165 @@
...
+    allPlayersScore = [];
+
+    gameState = GAME_STATE_PLAY_AGAIN;
+    return myOutputMessage;
+  }
+  if (gameState == GAME_STATE_PLAY_AGAIN) {
+    gameState = GAME_STATE_ROLL_DICE;
+    currentPlayer = 1;
+    myOutputMessage = main();
+  }
```
mahe993, line 163: Interesting, this is very good understanding of control flow. However, i wouldn't recommend this pattern because it is unnecessary complexity. We have 4 game modes, and within each we have a `if` checking and `return` within that `if`. That means that if code manages to reach this point , `gameState == GAME_STATE_PLAY_AGAIN` is implicitly true already.

Maybe we just need a default `myOutputMessage` and there would be no need for `myOutputMessage = main()`. Well done regardless! happy to see that your control flow is top tier now

mahe993:
Well done Marina, you really do pick up fast and I can see the code quality. Expecting more good stuff from you!

PR #576, opened 2024-02-14
27-3 Jia Hui

sinsinex:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

How many hours did you spend on this assignment? _At least 8 hours._

Please fill in one error and/or error message you received while working on this assignment. _Unable to commit properly to git, functions not working the way I intended_

What part of the assignment did you spend the most time on? _Trying to understand how I can get the game to run for a second time for the second player._

Comfort Level (1-5): 3 (assuming 1 being least comfortable - this was rather challenging for me)

Completeness Level (1-5): 3 (I hit the basics and I think that's good enough for me) 

What did you think of this deliverable? Definitely a step up from the scissors-paper-stone assignment, but still doable. 

Is there anything in this code that you feel pleased about? Figuring out how to combine the two numbers based on the user's input.

What's one aspect of your code you would like specific, elaborate feedback on? If there's better ways to refactor the code.

### DIFF
--- a/index.html
+++ b/index.html
@@ -18,7 +18,7 @@
       }
 
       #container {
-        background-color: pink;
+        background-color: rgb(235, 220, 252);
         margin: 40px auto;
         max-width: 800px;
         padding: 38px 31px;
@@ -49,7 +49,7 @@
   </head>
 
   <body>
-    <h1 id="header">Basics: Beat That! 🚀</h1>
+    <h1 id="header">Basics: Beat That! 🎲</h1>
     <div id="container">
       <p>Input:</p>
       <input id="input-field" />
@@ -57,6 +57,21 @@
       <button id="submit-button">Submit</button>
       <p>Output:</p>
       <div id="output-div"></div>
+    </div>
+    <style>
+      .centered-text {
+        text-align: center;
+      }
+    </style>
+    <div class="centered-text">
+      <p>Hello! Welcome to Beat That! 🎲</p>
+      <p>Grab a friend because this is a two-player game.👥</p>
+      <p>
+        The rules are simple. Roll two dice, and create a two-digit number from
+        the numbers rolled.
+      </p>
+      <p>The player with the highest number wins! 🏆</p>
+      <p>Ready to play? Player 1, click submit to start the game.</p>
     </div>
     <!-- Import program logic -->
     <script src="script.js"></script>
--- a/index.html
+++ b/index.html
@@ -62,16 +62,25 @@
       .centered-text {
         text-align: center;
       }
+      .hidden-text {
+        display: none;
+      }
     </style>
-    <div class="centered-text">
-      <p>Hello! Welcome to Beat That! 🎲</p>
-      <p>Grab a friend because this is a two-player game.👥</p>
-      <p>
+    <div id="gameInstructions" class="centered-text">
+      <p class="instruction-text">Hello! Welcome to Beat That! 🎲</p>
+      <p class="instruction-text">
+        Grab a friend because this is a two-player game.👥
+      </p>
+      <p class="instruction-text">
         The rules are simple. Roll two dice, and create a two-digit number from
         the numbers rolled.
       </p>
-      <p>The player with the highest number wins! 🏆</p>
-      <p>Ready to play? Player 1, click submit to start the game.</p>
+      <p class="instruction-text">
+        The player with the highest number wins! 🏆
+      </p>
+      <p class="instruction-text">
+        Ready to play? Player 1, click submit to start the game.
+      </p>
     </div>
     <!-- Import program logic -->
     <script src="script.js"></script>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -58,6 +58,30 @@ <h1 id="header">Basics: Beat That! 🚀</h1>
...
       <div id="output-div"></div>
     </div>
+    <style>
+      .centered-text {
+        text-align: center;
+      }
+      .hidden-text {
+        display: none;
+      }
+    </style>
```
floatingtales, line 68: great that you've added style tags!

But this belong within the `<header> header </header>` as opposed to the `<body> body </body>`

index.html
```diff
@@ -58,6 +58,30 @@ <h1 id="header">Basics: Beat That! 🚀</h1>
...
+        The rules are simple. Roll two dice, and create a two-digit number from
+        the numbers rolled.
+      </p>
+      <p class="instruction-text">
+        The player with the highest number wins! 🏆
+      </p>
+      <p class="instruction-text">
+        Ready to play? Player 1, click submit to start the game.
+      </p>
+    </div>
```
floatingtales, line 84: Great that you've added instructions here! Also great usage of classes and ids here!

Slight nitpick, id names usually have the same case styling as class names, they're in what we call `kebab-case`. So it might be better to name the id tag as `game-instructions`

index.html
```diff
@@ -74,6 +98,12 @@ <h1 id="header">Basics: Beat That! 🚀</h1>
 
         // Reset input value
         input.value = "";
+
+        // Hide game instructions text once the game starts
+        var instructions = document.querySelectorAll(".instruction-text");
+        instructions.forEach(function (instruction) {
+          instruction.classList.add("hidden-text");
+        });
```
floatingtales, line 106: Awesome job on doing this! You do know how to take in a tag with the `querySelector` and hide it! I don't particularly like the `array.forEach()` function because it's a bit finicky, but this works!

p.s. I'd like to use a for loop to iterate through this
```
for(var i = 0; i<instructions.length; i+=1){
  instructions[i].classList.add("hidden-text")
}
```

script.js
```diff
@@ -1,4 +1,141 @@
...
+//2 game modes, start with first player
+var modeRollDice = "modeRollDice";
+var modeChooseDiceOrder = "modeChooseDiceOrder";
+var currentPlayer = "Player 1";
+
+//Game starts with rolling of dice
+var gameMode = modeRollDice;
+
+var rollSingleDice = function () {
+  return Math.floor(Math.random() * 6) + 1;
```
floatingtales, line 29: nice way of making this a one-liner!

script.js
```diff
@@ -1,4 +1,141 @@
...
+var player2 = "Player 2";
+var player1DiceRoll = [];
+var player2DiceRoll = [];
+var player1Number;
+var player2Number;
+var player1WinCount = 0;
+var player2WinCount = 0;
+var player1LoseCount = 0;
+var player2LoseCount = 0;
+var drawCount = 0;
```
floatingtales, line 18: great usage of global variables here!

But there are slightly redundant variables here. because player1WinCount will always be exactly the same as player2LoseCount, and vice versa.

script.js
```diff
@@ -1,4 +1,141 @@
...
+
+//Game starts with rolling of dice
+var gameMode = modeRollDice;
+
+var rollSingleDice = function () {
+  return Math.floor(Math.random() * 6) + 1;
+};
+
+var rollDiceAndAssign = function () {
+  var getDiceNumber = [rollSingleDice(), rollSingleDice()];
```
floatingtales, line 33: This works for a 2 dice game. It might be a bit challenging if you want to expand this into having variable amount of dice tho!

script.js
```diff
@@ -1,4 +1,141 @@
...
+
+  if (currentPlayer === "Player 1") {
```
floatingtales, line 35: ahh yes! A `===` strict equality operator!

It actually is a more recommended way of doing as opposed to a `==` equality operator.

the only difference being the `===` checks the data type as well as the value. as opposed to a `==`.
i.e.
```
1=="1" //true
1==="1" //false
```

script.js
```diff
@@ -1,4 +1,141 @@
...
+Your dice numbers are: <br> 
+Dice 1: ${getDiceNumber[0]} <br>
+Dice 2: ${getDiceNumber[1]} <br> 
+Indicate your choice of order by entering 1 or 2 to assume the first value of the combined number.`;
+    return diceRollMessage;
+  }
+
+  //Combine numbers based on chosen order and show user combined number
+  if (gameMode === modeChooseDiceOrder) {
+    var orderChosen = Number(input);
```
floatingtales, line 97: this is why you need to *cast* the string into a number,  because of the strict equality!

However the way that you chose to do it might have a weird edge case that should be considered:

try inputting a Hex-value string like say `0x000001`

It will still run!

floatingtales (commented):
Great job Jia Hui! This is a very neat beat that project! I'm excited to see what you can do for the blackjack project!

PR #577, opened 2024-02-15
Basics Beat That Submission: Elky

elkyli:

🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

Over 10++ hours 🥲

Returned with an empty state as I was unable to change the game modes.

Figuring out how to compare the two scores and also pushing to Git. Took an hour plus to figure out pushing to Git as I had a few errors.

Comfort Level (1-5): 3

Completeness Level (1-5): 3

I would like to be able to spend more time to cover more scenarios (e.g. if user doesn't input 1 or 2, what message will they see)

At least I'm done with basics!

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title>Coding Basics</title>
+    <title>Beat That!</title>
     <style>
       * {
         box-sizing: border-box;
@@ -18,7 +18,7 @@
       }
 
       #container {
-        background-color: pink;
+        background-color: rgb(225, 251, 186);
         margin: 40px auto;
         max-width: 800px;
         padding: 38px 31px;
@@ -40,7 +40,7 @@
       }
 
       #output-div {
-        background-color: lightgrey;
+        background-color: rgb(254, 255, 243);
         margin: 20px auto;
         padding: 20px;
         width: 100%;
@@ -49,18 +49,27 @@
   </head>
 
   <body>
-    <h1 id="header">Basics: Beat That! 🚀</h1>
+    <h1 id="header">Welcome to Beat That 🎲</h1>
     <div id="container">
-      <p>Input:</p>
-      <input id="input-field" />
-      <br />
-      <button id="submit-button">Submit</button>
-      <p>Output:</p>
-      <div id="output-div"></div>
+      <body>
+          Create a two-digit number by selecting the order of your dice rolls.
+          <br />
+          The player with the highest number wins!
+        </p>
+
+        👇 Click <b>submit</b> to start the game.
+        <p>Input:</p>
+        <input id="input-field" />
+        <br />
+        <button id="submit-button">Submit</button>
+        <p>Current state:</p>
+        <div id="output-div"></div>
+      </body>
     </div>
     <!-- Import program logic -->
     <script src="script.js"></script>
     <!-- Define button click functionality -->
+
     <script>
       var button = document.querySelector("#submit-button");
       button.addEventListener("click", function () {
--- a/script.js
+++ b/script.js
@@ -1,4 +1,81 @@
+// Create game states
+var playerToRollDice = "rollDice";
+var playerChooseDiceOrder = "chooseDiceOrder";
+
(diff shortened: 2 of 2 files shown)


index.html
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title>Coding Basics</title>
+    <title>Beat That!</title>
```
floatingtales, line 4: Love that you changed the title here!

index.html
```diff
@@ -40,7 +40,7 @@
       }
 
       #output-div {
-        background-color: lightgrey;
+        background-color: rgb(254, 255, 243);
```
floatingtales, line 43: It's great that you're playing with rgb for the display! This adds a great deal of personal touch to your code!

index.html
```diff
@@ -49,18 +49,27 @@
...
+        </p>
+
+        👇 Click <b>submit</b> to start the game.
+        <p>Input:</p>
+        <input id="input-field" />
+        <br />
+        <button id="submit-button">Submit</button>
+        <p>Current state:</p>
+        <div id="output-div"></div>
+      </body>
```
floatingtales, line 67: Generally, you shouldn't put a `<body>` tag within another `<body>` tag. a standard `<p>` tag or a `<div>` tag would suffice here.

script.js
```diff
@@ -1,4 +1,81 @@
...
+
+var compareScores = "compareScores";
+
+// Default game state
+var currentGameMode = playerToRollDice;
+
+var player1DiceRoll1, player1DiceRoll2;
+var player2DiceRoll1, player2DiceRoll2;
+
+var player1DiceOrder, player2DiceOrder;
```
floatingtales, line 16: while this method of declaring multiple variables work, I'd much prefer if you would just have them line by line as opposed to them being in one line. It's generally more readable that way.

```
var player1DiceOrder
var player2DiceOrder
```

script.js
```diff
@@ -1,4 +1,81 @@
...
+var player2ToRollDice = "rollDice2";
+var player2ChooseDiceOrder = "chooseDiceOrder2";
+
+var compareScores = "compareScores";
+
+// Default game state
+var currentGameMode = playerToRollDice;
+
+var player1DiceRoll1, player1DiceRoll2;
+var player2DiceRoll1, player2DiceRoll2;
```
floatingtales, line 14: I'd probably explore the usage of an array here. Although this works, it gets harder to scale the game IF you decide to make a game with multiple dice like say, 5 dice, 6 dice.

you'd have to have the same number of variables with the dice. and it's not super ideal.

script.js
```diff
@@ -1,4 +1,81 @@
...
+    return `Player 1, here's what you rolled! 🎲<br><br> Dice 1: ${player1DiceRoll1} <br> Dice 2: ${player1DiceRoll2} <br><br> Choose your dice order by entering '1' or '2' in the input field.`;
+  }
+
+  // player 1 choose dice order
+  if (currentGameMode == playerChooseDiceOrder) {
+    if (input == "1") {
+      player1DiceOrder = "" + player1DiceRoll1 + player1DiceRoll2;
+    } else {
+      player1DiceOrder = "" + player1DiceRoll2 + player1DiceRoll1;
+    }
```
floatingtales, line 40: This is a novel way of combining the numbers! adding a string to a number would cast that number as a string as opposed to a number, hence why this works.

Also, it might be best to add an input validation somewhere here as this would mean that if you'd have an input other than `1` it will always default to the second choice. i.e. if let's say my input is `3` it'll go to the else block, even when my input should not be a valid input like say `"what's up"` it will still go to the else block.

floatingtales (commented):
Hey Elky! This is a great attempt on the second project! I know that you tried your best with this, but here's a couple of notes:

- Familiarize yourself with arrays, I don't see that you're using any arrays here. There are cases in which it's preferable to use arrays than not here.
- The main function is rather slightly bulky. Maybe it's best to have smaller helper functions to help you alleviate the bulk from the main function.

All in all, it's a great attempt at the beat that project and I am excited to see what you can do in your blackjack project!

